PR #16, opened 2016-02-28
Fix: SyntaxKind code checks, update TS peerDependency ^1.7.3 (fixes #15)

### LINKED ISSUE #15: TypeScript upgrade breaks the world
It looks like there were some changes in the TypeScript output between 1.6.x and 1.8.x that are causing identifiers to be labeled as keywords in the output. As such, most tests are failing right now.

I'm looking into this.

JamesHenry:
Fixes #15

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -293,11 +293,11 @@ function getTokenType(token) {
         }
     }
 
-    if (token.kind >= 68 && token.kind <= 112) {
+    if (token.kind >= SyntaxKind.FirstKeyword && token.kind <= SyntaxKind.LastFutureReservedWord) {
         return "Keyword";
     }
 
-    if (token.kind >= 15 && token.kind <= 66) {
+    if (token.kind >= SyntaxKind.FirstPunctuation && token.kind <= SyntaxKind.LastBinaryOperator) {
         return "Punctuator";
     }
 
--- a/package.json
+++ b/package.json
@@ -58,6 +58,6 @@
     "object-assign": "^4.0.1"
   },
   "peerDependencies": {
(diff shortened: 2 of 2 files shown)


lib/ast-converter.js
```diff
@@ -293,11 +293,11 @@ function getTokenType(token) {
         }
     }
 
-    if (token.kind >= 68 && token.kind <= 112) {
+    if (token.kind >= 70 && token.kind <= 114) {
```
nzakas: Rather than just hardcoding the numbers, I think we should try to make use of the token boundary values: https://github.com/Microsoft/TypeScript/commit/b24c3a7aff3bf6e50a43256c72ae7a6851e35673#diff-29796169c2bd635765a9f574727f77f3R310

As far as I can tell, using `SyntaxKind.FirstKeyword` and `SyntaxKind.LastFutureReservedWord` here might inoculate us from this problem in the future.

lib/ast-converter.js
```diff
@@ -293,11 +293,11 @@ function getTokenType(token) {
...
         return "Keyword";
     }
 
-    if (token.kind >= 15 && token.kind <= 66) {
+    if (token.kind >= 15 && token.kind <= 68) {
```
nzakas: Same here with `SyntaxKind.LastBinaryOperator`

JamesHenry:
Thanks for the feedback and for setting up travis - I updated the PR.

It seems like it failed on node 0.10.x build, but passed on the others. NPM version issue?

nzakas:
Looks like the TypeScript module isn't being installed because it's a peer dependency. We can probably stop testing in 0.10 at this point, so I'm not too worried.

PR #17, opened 2016-03-04
Fix: Ensure true, false and null values are not stringifed (fixes #1)

### LINKED ISSUE #1: TypeError being thrown in no-octal-escape rule
Hi @nzakas,

Thanks so much for getting the ball rolling on this!

I have really enjoyed (and benefitted from) using ESLint with a large Babel-transpiled codebase and it has been flawless. That codebase is now being migrated to TypeScript and I really hope to continue using ESLint, so you may see some issues on here from me.

As a newcomer to world of JS ASTs and linter design, it may take a little while before I am actively contributing, but for now hopefully I can add value by simply reporting what I find.

(Note: I am using ESLint via eslint-loader for webpack)
ESLint: 1.9.0
eslint-loader: 1.1.1
typescript-eslint-parser: 0.1.0-alpha-1

My first issue is that the following error is being thrown a lot within `no-octal-escape.js`:

```
Module build failed: TypeError: Cannot read property 'match' of undefined
    at EventEmitter.Literal ({{PROJECT_PATH}}/node_modules/eslint/lib/rules/no-octal-escape.js:26:33)
```

- The code in `no-octal-escape` responsible for the TypeError:

```
module.exports = function(context) {

    return {

        "Literal": function(node) {
            if (typeof node.value !== "string") {
                return;
            }

            var match = node.raw.match(/^([^\\]|\\[^0-7])*\\([0-3][0-7]{1,2}|[4-7][0-7]|[0-7])/),
                octalDigit;

            if (match) {
                octalDigit = match[2];

                // \0 is actually not considered an octal
                if (match[2] !== "0" || typeof match[3] !== "undefined") {
                    context.report(node, "Don't use octal: '\\{{octalDigit}}'. Use '\\u....' instead.",
                            { octalDigit: octalDigit });
                }
            }
        }

    };

};
```

- I tracked down the first "offender" to be this call to angular bootstrap, and specifically the ` strictDi: true`

```
angular.bootstrap( document, [ 'app' ], {
   strictDi: true
})
```

- The `node` that gets passed through to the `"Literal": function(node) {` handler (by the typescript-eslint-parser?) has a `value` of `'true'` so it is passing the typeof check, but it does not have a `raw` property at all.

E.g.

```
{ type: 'Literal',
  range: [ 1890, 1894 ],
  loc:
   { start: { line: 91, column: 12 },
     end: { line: 91, column: 16 } },
  value: 'true',
  parent:
   { type: 'Property',
     range: [ 1880, 1894 ],
     key:
      { type: 'Identifier',
        range: [Object],
        loc: [Object],
        name: 'strictDi',
        parent: [Circular] },
     value: [Circular],
     computed: false,
     method: false,
     shorthand: false,
     kind: 'init',
     parent:
      { type: 'ObjectExpression',
        range: [Object],
        loc: [Object],
        properties: [Object],
        parent: [Object] } } }
```

- My guess is that node.value should be the boolean value `true` rather than the string `'true'`.

Please let me know if you need any more info!

JamesHenry:
Fixes #1

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1346,21 +1346,21 @@ module.exports = function(ast, extra) {
             case SyntaxKind.TrueKeyword:
                 assign(result, {
                     type: "Literal",
-                    value: "true"
+                    value: true
                 });
                 break;
 
             case SyntaxKind.FalseKeyword:
                 assign(result, {
                     type: "Literal",
-                    value: "false"
+                    value: false
                 });
                 break;
 
             case SyntaxKind.NullKeyword:
                 assign(result, {
                     type: "Literal",
-                    value: "null"
+                    value: null
                 });
                 break;
 


JamesHenry:
Same Node 0.10.x build failure :+1:

nzakas:
I updated the Travis build to not check Node 0.10.

PR #23, opened 2016-03-11
Dependencies update (TS 1.8.7)

valorkin:
- all dependencies updated
- deps made immutable (save-exact)
- typescript added as dev dependency
- TS version set to 1.8.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,18 +26,19 @@
     }
   ],
   "devDependencies": {
-    "chai": "^1.10.0",
-    "dateformat": "^1.0.11",
-    "eslint": "^2.2.0",
-    "eslint-config-eslint": "^3.0.0",
-    "eslint-release": "^0.3.0",
-    "istanbul": "~0.2.6",
-    "leche": "^1.0.1",
-    "mocha": "^2.0.1",
-    "npm-license": "^0.2.3",
-    "semver": "^4.1.1",
-    "shelljs": "^0.3.0",
-    "shelljs-nodecli": "^0.1.1"
+    "chai": "3.5.0",
+    "dateformat": "1.0.11",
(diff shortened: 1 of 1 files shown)


nzakas:
Thanks. We follow the standard ESLint guidelines for pull requests:
http://eslint.org/docs/developer-guide/contributing/pull-requests

So, we need separate PRs for each commit and for the commit summaries to be in a specific format.

valorkin:
@nzakas ready for merge

package.json
```diff
@@ -26,18 +26,19 @@
     }
   ],
   "devDependencies": {
-    "chai": "^1.10.0",
```
nzakas: I'd prefer not to hardcode dependency versions.

package.json
```diff
@@ -26,18 +26,19 @@
...
+    "eslint-config-eslint": "3.0.0",
+    "eslint-release": "0.3.0",
+    "istanbul": "0.4.2",
+    "leche": "2.1.1",
+    "mocha": "2.0.1",
+    "npm-license": "0.3.1",
+    "semver": "5.1.0",
+    "shelljs": "0.6.0",
+    "shelljs-nodecli": "0.1.1",
+    "typescript": "1.8.7"
```
nzakas: Why was this added? It's already in peerDependencies.

package.json
valorkin: you prefer to have broken version because of someone changes?

package.json
valorkin: using npm 3 it is not getting installed for development
and not changing npm module
so it is nice to have

package.json
nzakas: Please revert these. Thanks.

PR #28, opened 2016-03-13
New: Implements JSX syntax (fixes #18)

### LINKED ISSUE #18: Implement JSX syntax
Phase 2 is to add JSX syntax support into this parser. This would involve:

1. Converting the TypeScript JSX syntax tree into an ESTree-compatible one.
1. Convert the TypeScript tokens into Esprima-style tokens.
1. Making the result pass the JSX tests that are already tests in tests/fixtures/ecma-features/jsx.

@JamesHenry are you interested in taking a stab at this?

JamesHenry:
@nzakas would be great to get your feedback on this!

**Note**: I added `lodash.unescape` as a dependency to ensure that HTML entities in JSX tags are decoded appropriately. Now that lodash is entirely modular, I felt that it was more prudent to include a micro-dependency than reinvent the wheel.

### DIFF
--- a/tests/lib/ecma-features.js
+++ b/tests/lib/ecma-features.js
@@ -43,12 +43,24 @@ var assert = require("chai").assert,
 var FIXTURES_DIR = "./tests/fixtures/ecma-features";
 // var FIXTURES_MIX_DIR = "./tests/fixtures/ecma-features-mix";
 
+var filesWithOutsandingTSIssues = [
+    "jsx/embedded-tags", // https://github.com/Microsoft/TypeScript/issues/7410
+    "jsx/namespaced-attribute-and-value-inserted", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/namespaced-name-and-attribute", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/test-content", // https://github.com/Microsoft/TypeScript/issues/7471
+    "jsx/multiple-blank-spaces"
+];
+
 var testFiles = shelljs.find(FIXTURES_DIR).filter(function(filename) {
     return filename.indexOf(".src.js") > -1;
+}).filter(function(filename) {
+    return filesWithOutsandingTSIssues.every(function(fileName) {
+        return filename.indexOf(fileName) === -1;
+    });
 }).map(function(filename) {
     return filename.substring(FIXTURES_DIR.length - 1, filename.length - 7);  // strip off ".src.js"
 }).filter(function(filename) {
-    return !(/jsx|error\-|invalid\-|globalReturn|experimental|newTarget/.test(filename));
+    return !(/error\-|invalid\-|globalReturn|experimental|newTarget/.test(filename));
 });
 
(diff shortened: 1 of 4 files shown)


tests/lib/ecma-features.js
```diff
@@ -43,12 +43,24 @@ var assert = require("chai").assert,
 var FIXTURES_DIR = "./tests/fixtures/ecma-features";
 // var FIXTURES_MIX_DIR = "./tests/fixtures/ecma-features-mix";
 
+var filesWithOutsandingTSIssues = [
+    "jsx/embedded-tags", // https://github.com/Microsoft/TypeScript/issues/7410
+    "jsx/namespaced-attribute-and-value-inserted", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/namespaced-name-and-attribute", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/test-content", // https://github.com/Microsoft/TypeScript/issues/7471
+    "jsx/multiple-blank-spaces"
```
JamesHenry: I have not had chance to report this one yet, but it is almost certainly another bug with some internal TypeScript utilities. The body of the AST is correct, but we cannot currently populate tokens array properly because of the issue.

nzakas:
This looks really good. The only thing I think is missing is that it appears the parser will always parse JSX syntax. We may want to ensure that `parse()` only parses JSX when `ecmaFeatures.jsx` is true (as I'm sure there are people using TypeScript without JSX).

JamesHenry:
Sorry if I am missing something obvious, but wouldn't the changes I made in `parser.js` (https://github.com/eslint/typescript-eslint-parser/pull/28/files#diff-09461573a85e2d94f056dd6814769042R112) achieve exactly that?

You can try creating and running an `index.js` in the root of the project with the following content as a quick POC:

```
var code = '<div>@test content</div>';
var parser = require('./parser').parse
console.log(parser(code, {ecmaFeatures: {jsx: true}}))
```

If you set anything other than `jsx: true`, it will throw.

Again, if I am missing something else, please let me know and I will fix it!

nzakas:
All that ends up doing is changing the filename passed at https://github.com/eslint/typescript-eslint-parser/pull/28/files#diff-09461573a85e2d94f056dd6814769042R117

I'm not sure changing that filename is the same as enabling/disabling JSX support. Are you?

JamesHenry:
But... https://github.com/eslint/typescript-eslint-parser/pull/28/files#diff-09461573a85e2d94f056dd6814769042R154? :)

As well using it to determine that filename, we are using `extra.ecmaFeatures.jsx` to inform the setting of the jsx option on the TypeScript compiler, and as we know from: https://github.com/Microsoft/TypeScript/wiki/JSX#basic-usage

> In order to use JSX you must do two things.
>
> - Name your files with the .tsx extension
> - Enable the jsx option

...so that is both aspects covered, right?

nzakas:
Oh, oops! Sorry, my Lyme brain must be getting to me. This looks good!

nzakas:
I just pushed a new version with this fix.

PR #32, opened 2016-03-21
Fix: Remove console logging of node.kind (fixes #29)

### LINKED ISSUE #29: convert() logging node.kind to console
The `default` on `convert`'s `switch` on `node.kind` [logs it to the console][node-kind-log]. I understand that may be useful while hacking on the parser, but it's kinda distracting when trying to actually *use* the parser. :-)

Maybe this could be hidden behind a debug variable or configuration option?

[node-kind-log]: https://github.com/eslint/typescript-eslint-parser/blob/master/lib/ast-converter.js#L1378

tomxtobin:
This was probably left behind during development as per discussion in
issue #29.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1552,7 +1552,6 @@ module.exports = function(ast, extra) {
                 return convert(node.expression, parent);
 
             default:
-                console.log("unsupported node.kind:", node.kind);
                 result = null;
         }
 


tomxtobin:
I complied with @eslintbot's requests. ;-)

nzakas:
Thanks!

PR #49, opened 2016-07-27
New: Attaches comments to the ESTree AST (fixes #31)

### LINKED ISSUE #31: Phase 3: Attach comments
Phase 3 is to add attaching leading/trailing comments to the AST inside the parser. This would involve:

1. Retrieving comments for each node.
1. Converting the TypeScript comments into Esprima-style comments.
1. Adding the comments into `leadingComments` and `trailingComments` arrays on appropriate nodes.
1. Making the result pass the tests that are already tests in tests/fixtures/attach-comments.

JamesHenry:
See https://github.com/eslint/typescript-eslint-parser/pull/34 for all the discussion leading up to this implementation


lib/ast-converter.js
```diff
@@ -268,6 +267,8 @@ function getTokenType(token) {
 
             case SyntaxKind.GetKeyword:
             case SyntaxKind.SetKeyword:
+            case SyntaxKind.TypeKeyword:
```
nzakas: It looks like there are a bunch of unrelated changes in this file?

JamesHenry: They surfaced as bugs/missing checks when implementing the original work on this branch. I left them in because they seem like bugs waiting to happen, but I am happy to remove them if you feel that would be cleaner.

nzakas:
As noted inline, it looks like there are a lot of unrelated changes in this PR. Can you narrow it down so the changes are only related to finding the comments?

JamesHenry:
It is not as simple as saying they are unrelated to finding the comments... For example:
```
case SyntaxKind.TrueKeyword:
                assign(result, {
                    type: "Literal",
                    value: true,
                    raw: "true" // <- This was added in the PR
                });
                break;
```

The addition of the raw values for TrueKeyword, FalseKeyword and NullKeyword is necessary to make the attach-comments tests pass.

nzakas:
Oh I see, thanks for explaining. In that case, I think this is good to go.

JamesHenry:
Great, thanks!

PR #53, opened 2016-08-09
New: Class implements generic syntax (fixes #44)

### LINKED ISSUE #44: Add typed heritage parameters to class conversion
It's possible to do things like:

```ts
class Foo implements Bar<S> {

}
```

Right now, we completely lose the `S` part of the conversion. We should mimic what TypeScript does to remedy this.

JamesHenry:


nzakas:
This looks good. I'm assuming it still needs to be updated to ignore the leading whitespace in the location information?

JamesHenry:
No, that is in this PR :) https://github.com/JamesHenry/typescript-eslint-parser/blob/e2f854539e2fd50f9bae5b88d1a625d6c016cf76/lib/ast-converter.js#L472

JamesHenry:
And in the `tokens: []` you can see the `,` Punctuator ends on 27, but the `T` Identifier doesn't start until 28

PR #59, opened 2016-08-20
Fix: UpdateExpression detection and operator format (fixes #58)

### LINKED ISSUE #58: i++ should be parsed as UpdateExpression/++ rather than UnaryExpression/PlusPlusToken
I ran into this while debugging why this got flagged with eslint:

```ts
let i = 0;
function f() {
  i++;
}
f();
```

The error is `'i' is never reassigned. Use 'const' instead`.

The esprima parse for the `i++` looks like this:

```json
          {
            "expression": {
              "argument": {
                "name": "i",
                "type": "Identifier"
              },
              "operator": "++",
              "prefix": false,
              "type": "UpdateExpression"
            }
```

whereas the typescript-eslint-parser equivalent is:

```json
            "expression": {
              "argument": {
                "name": "i",
                "type": "Identifier"
              },
              "operator": "PlusPlusToken",
              "prefix": false,
              "type": "UnaryExpression"
            }
```

JamesHenry:


nzakas:
Can you add a test for this change?

JamesHenry:
There didn't seem to be an obvious place for such a fundamental assertion to live, so I have created a new "basics" directory in fixtures.

Just let me know if you would prefer it was organised differently

nzakas:
That's fine for now -- I want to reorganize at some point, but unsure the best way to do that atm.

Thanks!

PR #72, opened 2016-08-26
New: Add property decorators to AST (fixes #71)

### LINKED ISSUE #71: Add property decorators to AST
As per #63

JamesHenry:
We were previously just stripping `ClassProperty` nodes entirely. This PR adds support for them, and their decorators


nzakas:
Lgtm. Btw, you can just reference the master issue instead of creating new issues that don't say anything. Just use "refs" instead of "fixes".

PR #73, opened 2016-08-28
New: Add accessor decorators to AST (fixes #63)

### LINKED ISSUE #63: Milestone: Decorator Support
This issue it to track progress against supporting decorators within the final AST that the parser produces. We currently remove them during parsing.

We need to ensure that the following are accurately carried over from the original TypeScript AST:

- [x] Parameter Decorators
 - [x] On instance member
 - [x] On static member
 - [x] In constructor
- [x] Method Decorators
 - [x] On instance member
 - [x] On static member
- [x] Accessor Decorators
 - [x] On instance member
 - [x] On static member
- [x] Property Decorators
 - [x] On instance member
 - [x] On static member
- [x] Class Decorators

We also need to account for the difference between decorators and decorator factories:

**Decorator:**
unspecified-11

**Decorator Factory:**
unspecified-12

At the AST-level, the decorator node has an `Identifier` TSNode as an `expression`, whereas the decorator factory has a `CallExpression` TSNode.

Decorators are not currently covered by the ESTree spec as they are still a fairly early-stage es-proposal, but Angular 2, for example, has adopted them as a core aspect of the framework.

TypeScript adds a `decorators` array to the node in question, and I propose we do the same. 

Resources:
https://www.typescriptlang.org/docs/handbook/decorators.html

JamesHenry:
Added tests for accessor decorators/decorator factories.

No further logic was needed as ESTree getters and setters are just different `kind`s of `MethodDefinition`s.

### DIFF
--- a/tests/fixtures/typescript/decorators/accessor-decorators/accessor-decorator-factory-instance-member.result.js
+++ b/tests/fixtures/typescript/decorators/accessor-decorators/accessor-decorator-factory-instance-member.result.js
@@ -1,0 +1,645 @@
+module.exports = {
+    "type": "Program",
+    "range": [
+        0,
+        72
+    ],
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 4,
+            "column": 1
+        }
+    },
+    "body": [
+        {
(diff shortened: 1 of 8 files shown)


nzakas:
Awesome! LGTM

PR #86, opened 2016-09-10
Fix: Convert TSTypeOfExpression to UnaryExpression (fixes #85)

### LINKED ISSUE #85: `typeof` falling through to auto-conversion
We are currently not explicitly handling a standard `typeof` statement

E.g.
```ts
typeof 'str'
```

It should be converted to a `UnaryExpression`, currently being auto-converted to a `TSTypeOfExpression`

JamesHenry:


nzakas:
Nice catch. LGTM

PR #88, opened 2016-09-15
New: Accessibility Modifiers (fixes #87)

### LINKED ISSUE #87: ClassProperty and MethodDeclaration Visibility Declarations
2.0.2

0.2.0

```
class Animal {
  public name: string
}
```

I expected an `ASTNode` for `ClassProperty` with `public` represented on it.

**What happened?**
I had to use [`context.getText(node)`](https://github.com/dannyfritz/eslint-plugin-typescript/blob/5c9bff1197fb65f766f5f18020f99c23bb71aa23/lib/rules/member-access.js#L49) to do a string find for the [visibility declaration](https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers). I'm wondering if I either missed something or if we are missing it in the conversion.

---

References this PR on the `typescript-plugin` https://github.com/nzakas/eslint-plugin-typescript/pull/4#issuecomment-246378378

dannyfritz:
Issue (#87)

This pull request is about getting the `public`, `protected`, and `private` keywords into the ESLint AST.

The accessibility modifier keywords can appear on `ClassProperty` and `MethodProperty`.

http://astexplorer.net/#/FfIL8GurdC

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -499,6 +499,32 @@ module.exports = function(ast, extra) {
         }
 
         /**
+         * Gets a TSNode's accessibility level
+         * @param {TSNode} tsNode The TSNode
+         * @returns {string | null} accessibility "public", "protected", "private", or null
+         */
+        function getTSNodeAccessibility(tsNode) {
+            var modifiers = tsNode.modifiers;
+            if (!modifiers) {
+                return null;
+            }
+            for (var i = 0; i < modifiers.length; i++) {
+                var modifier = modifiers[i];
+                switch (modifier.kind) {
+                    case SyntaxKind.PublicKeyword:
+                        return "public";
+                    case SyntaxKind.ProtectedKeyword:
+                        return "protected";
+                    case SyntaxKind.PrivateKeyword:
+                        return "private";
+                    default:
+                        continue;
+                }
+            }
+            return null;
+        }
+
+        /**
          * Converts a TSNode's typeParameters array to a flow-like TypeParameterDeclaration node
          * @param {TSNode[]} typeParameters TSNode typeParameters
          * @returns {TypeParameterDeclaration} TypeParameterDeclaration node
@@ -984,6 +1010,7 @@ module.exports = function(ast, extra) {
                     value: convertChild(node.initializer),
                     computed: (node.name.kind === SyntaxKind.ComputedPropertyName),
                     static: Boolean(node.flags & ts.NodeFlags.Static),
+                    accessibility: getTSNodeAccessibility(node),
                     decorators: (node.decorators) ? node.decorators.map(function(d) {
                         return convertChild(d.expression);
                     }) : []
@@ -1066,6 +1093,7 @@ module.exports = function(ast, extra) {
                         computed: isMethodNameComputed,
                         static: Boolean(node.flags & ts.NodeFlags.Static),
                         kind: "method",
+                        accessibility: getTSNodeAccessibility(node),
                         decorators: (node.decorators) ? node.decorators.map(function(d) {
                             return convertChild(d.expression);
                         }) : []
@@ -1157,6 +1185,7 @@ module.exports = function(ast, extra) {
                     key: constructorKey,
                     value: constructor,
                     computed: constructorIsComputed,
+                    accessibility: getTSNodeAccessibility(node),
                     static: constructorIsStatic,
                     kind: (constructorIsStatic || constructorIsComputed) ? "method" : "constructor"
                 });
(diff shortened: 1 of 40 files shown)


lib/ast-converter.js
```diff
@@ -499,6 +501,31 @@ module.exports = function(ast, extra) {
         }
 
         /**
+         * Gets a TSNode's accessibility level
+         * @returns {string} accessibility "public", "protected", or "private"
+         */
+        function getTSNodeAccessibility() {
```
JamesHenry: It's a minor point, but I would usually prefer that TSNode be passed into the helper function so that it is a bit more portable.

I'm sure we'll be refactoring some of this code once the project is a bit more mature and things like this will make that process easier.

tests/fixtures/typescript/basics/class-with-accessibility-modifiers.src.ts
```diff
@@ -0,0 +1,9 @@
+class Foo {
+  private bar : string;
+  public getBar () {
+    return this.bar;
+  }
+  protected setBar (bar : string) {
+    this.bar = bar;
+  }
+}
```
JamesHenry: Please add at least one example of methods and properties which have multiple TS modifiers on them:

e.g. 

```ts
class X {
  public static yo: string = 'Init';
  protected get bye() {
    return 'bye';
  }
}
```

lib/ast-converter.js
```diff
@@ -420,6 +420,8 @@ module.exports = function(ast, extra) {
             return null;
         }
 
+        var accessibility;
```
dannyfritz: I really don't like this line. I'm going to have it always set the `accessibility` and just have it default to `null` instead to clean up code.

dannyfritz:
I'm noticing `constructor` should default to `accessibility: "public"`.

Also, there is a shorthand in TS for [specifying members in the `constructor` parameters](https://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties). Are type annotations parsed into the tree yet?

JamesHenry:
> I'm noticing constructor should default to accessibility: "public" and currently does not

I'm not sure what you mean by this. Can you please point me to an example?

Everything in a TypeScript class is public by default (just like a standard ES2015 class), we will only worry about adding `accessibility: "public"` in the AST for those nodes which have a PublicKeyword as a modifier.

JamesHenry:
> Also, there is a shorthand in TS for specifying members in the constructor parameters. Are type annotations parsed into the tree yet?

Yes, we will need to handle this behaviour in a separate PR as it will require some discussion. Thanks for flagging though!

dannyfritz:
Alright, if we're only concerned with explicit accessibility keywords then this should be good.

JamesHenry:
RE: The constructor:

Oh, I missed this comment from you:

> I really don't like this line. I'm going to have it always set the accessibility and just have it default to null instead to clean up code.

...and I can see you have already updated all the relevant existing ASTs.

@nzakas What do you think about this change?

If we do go with Danny's latest changes which adds "accessibility" to everything, then constructor should probably have it too

dannyfritz:
@JamesHenry https://github.com/eslint/typescript-eslint-parser/issues/87#issuecomment-247083910

JamesHenry:
Ok great, sorry I missed that discussion earlier in the week as I was travelling

JamesHenry:
So, to summarise, I still thinks it makes sense to keep the `accessibility` value to be null, unless the modifiers are explicitly used. They will only be relevant to specific plugin rules anyway

JamesHenry:
This LGTM now!

I'll leave it to @nzakas to merge to double check he is happy with "accessibility" over "visibility". 

We usually try and match existing ESTree or Flow constructs wherever possible, but these modifiers do not exist on classes for them yet. I therefore think it makes sense to go with TypeScript's term of "accessibility", as in this PR.

Thanks again, @dannyfritz!

dannyfritz:
I'm going to add accessibility to the constructor too. null still if nothing is specified. Then i think this PR is ready.

nzakas:
> So, to summarise, I still thinks it makes sense to keep the accessibility value to be null, unless the modifiers are explicitly used. 

Totally agree. 

> I'll leave it to @nzakas to merge to double check he is happy with "accessibility" over "visibility".

Eh, I don't have the energy to argue this point, so it's fine. :)

@dannyfritz it looks like you're still planning on making changes? Can you let us know when you do?

dannyfritz:
Alright. This PR is ready for review.

JamesHenry:
Thanks, @dannyfritz! This still LGTM.

PR #57, opened 2016-08-18
Update: Added travis ci support for modern node versions

amilajack:

PR #98, opened 2016-09-21
Build: Add Node 6.x to Travis

dannyfritz:
Finishing what #57 started.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,9 @@ sudo: false
 node_js:
     - "0.12"
     - 4
+    - 6
+before_script:
+    - npm install -g npm-install-peers
+    - npm-install-peers
 after_success:
     - npm run coveralls


.travis.yml
```diff
@@ -3,5 +3,8 @@ sudo: false
 node_js:
     - "0.12"
     - 4
+    - 6
+before_script:
+    - npm install typescript@^1.7.3
```
nzakas: Any way we can pull the version from package.json so we don't have to worry about keeping the versions in sync?

dannyfritz: We could add this as a `devDependency` and run it instead. https://www.npmjs.com/package/npm-install-peers

nzakas:
Also, this should be "Build:"

dannyfritz:
Oh geez, 0.12 can't run `npm-install-peers`:

```sh
const chalk = require('chalk')
^^^^^
SyntaxError: Use of const in strict mode.
```

dannyfritz:
Working with `npm-install-peers` to resolve this: https://github.com/spatie/npm-install-peers/issues/2

JamesHenry:
@nzakas I am wondering if node 0.12 should even be a consideration at this point... It's maintenance support will be discontinued at the end of the year, and node 4.x has long since been the LTS.

It would really surprise me if there are many users of TypeScript and node 0.12 as a combination out there.

What do you think?

dannyfritz:
Got a PR in for `npm-install-peers` and this should be good to go now.

https://github.com/spatie/npm-install-peers/pull/3

PR #100, opened 2016-09-22
Fix: Add loc and range data to generated VariableDeclarator node

JamesHenry:
I am really sorry this has now taken a 3rd PR to get right.

This adds `loc` and `range` data for the generated `VariableDeclarator` - core `indent` is currently throwing an error because `loc` is undefined


nzakas:
Not a problem, that's why we're still in experimental mode. :)

PR #104, opened 2016-10-13
Fix: Added missing yml lf gitattribute

JamesHenry:
Since the gitattribute file was added in https://github.com/eslint/typescript-eslint-parser/pull/97, my mac has automatically had unstaged changes relating to yml files.

I have never used gitattributes before, but adding a definition for yml seems to do the trick...

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 * text=auto
 *.js eol=lf
 *.ts eol=lf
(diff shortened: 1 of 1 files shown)


dannyfritz:
Hrmm. I wonder if `text=auto` was causing your mac to use `\r\n` or something really weird like that. My impression is it will choose the system default which should b fine.

Issue #113, opened 2016-11-09
Release build broken


nzakas:
When I tried to do a release, it errors out. I think this has to do with the `.gitattributes` change, but I'm unfamiliar with how that works.

Relevant log snip:

```
Running tests
Calculating changes for release
Release is 1.0.0
Generating changelog
Committing to git

[master 52c116a] Build: package.json and changelog update for 1.0.0
 1 file changed, 9 insertions(+)

Generating 1.0.0
npm ERR! Linux 3.16.0-4-amd64
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "version" "1.0.0"
npm ERR! node v4.4.7
npm ERR! npm  v2.15.8

npm ERR! Git working directory not clean.
npm ERR! M .travis.yml
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /var/lib/jenkins/workspace/Releases/typescript-eslint-parser Release/typescript-eslint-parser/npm-debug.log


npm ERR! Linux 3.16.0-4-amd64
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "run" "ci-release"
npm ERR! node v4.4.7
npm ERR! npm  v2.15.8
npm ERR! code ELIFECYCLE
npm ERR! typescript-eslint-parser@0.4.0 ci-release: `eslint-ci-release`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the typescript-eslint-parser@0.4.0 ci-release script 'eslint-ci-release'.
npm ERR! This is most likely a problem with the typescript-eslint-parser package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     eslint-ci-release
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs typescript-eslint-parser
npm ERR! Or if that isn't available, you can get their info via:
npm ERR! 
npm ERR!     npm owner ls typescript-eslint-parser
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /var/lib/jenkins/workspace/Releases/typescript-eslint-parser Release/typescript-eslint-parser/npm-debug.log
Build step 'Execute shell' marked build as failure
[ssh-agent] Stopped.
Finished: FAILURE
```

[...]

PR #114, opened 2016-11-09
Fix: Unset text attribute .travis.yml instead of using eol

JamesHenry:


platinumazure:
Is the issue that `*.yml` doesn't match `.travis.yml` due to the leading dot? Or is it specifically that the Travis file needs to be regarded as non-text?

JamesHenry:
@platinumazure the only known context is here https://github.com/eslint/typescript-eslint-parser/issues/113.

The gitattributes file was originally added to help by a windows user, to help windows users.

Originally the file contained just:
```
 * text=auto
  *.js eol=lf
  *.ts eol=lf
```

But then whenever I opened the project on my mac, I would get unstaged changes on the .travis.yml file, despite no actual diff/changes.

Adding the `*.yml eol=lf` was the original way to fix that - but it _seems_ to have had the side-effect of breaking the release job on jenkins... Which I am guessing is on a linux machine @nzakas?

platinumazure:
Just to get this out of the way, is there anything in the YAML spec which requires that linefeeds be CRLF or LF only?

platinumazure:
Okay, I've done some experimenting and I am fairly confident `*.yml` matches `.travis.yml`.

I think what's going on is the .travis.yml file in the index has CRLF line endings, despite the gitattributes policy specified for *.yml files. As is noted in the gitattributes docs, adding the attribute to gitattributes does not retroactively apply to files that were committed with the wrong LF pattern before the attribute was added. So I guess the PR I would rather see is the result of these operations:

1. Revert the .gitattributes change (so that `*.yml eol=lf` is the last line again)
1. Run `rm .travis.yml`
1. Run `git reset`
1. Run `git add -u`

Hopefully this means that the file will be re-committed with LF changes.

ilyavolodin:
@JamesHenry This wouldn't be the fix, since our release script is ran on Jenkins server, not Travis. So it's very unlikely this change will affect anything.

JamesHenry:
@ilyavolodin do you agree with @platinumazure's proposal?

JamesHenry:
I still don't understand why this is an issue on Jenkins, but not on Travis or locally...

JamesHenry:
But @platinumazure suggestion seems reasonable to me :)

ilyavolodin:
@JamesHenry Sorry, I think I misunderstood the core issue. I think it's worth a try at least. I can merge this in, and try running a release.

platinumazure:
As noted on the chat, I'm okay with merging this in as is just to try to unblock the release. I still hope someone will try [my suggestion here](https://github.com/eslint/typescript-eslint-parser/pull/114#issuecomment-259547458) at some point in the future. :smile:

ilyavolodin:
@nzakas I'm not going to run this release, since this includes a breaking change, and release script will automatically release this as v1.0.0. I'm leaving it up to you to decide if this is ready to be 1.0 release.

JamesHenry:
+1, as agreed on chat, we want to make sure we are all on the same page as to what release this will be

nzakas:
The release will be 1.0.0, that's fine. I think the project is probably solid enough to use with a few caveats.

@JamesHenry I'll leave it up to you which path to take to fix this, and then whoever is around can trigger the release.

Please be sure to update the commit message to include the issue (#113).

JamesHenry:
Thanks, @nzakas!

In which case, I will close this and submit a PR implementing @platinumazure's proposal.

PR #120, opened 2016-12-05
fix running without options

flying-sheep:
check out the [single change](https://github.com/eslint/typescript-eslint-parser/pull/120/files?w=1) by disabling whitespace diffing


tests/lib/parse.js
```diff
@@ -33,10 +33,19 @@ function getRaw(ast) {
 
 describe("parse()", function() {
 
+
+    describe("basic functionality", function() {
+
+        it("should parse an empty string", function() {
+            parser.parse("");
```
JamesHenry: Please add simple assertions for the resulting ASTs - they will be small enough to keep inline. I would prefer to be explicit rather than the successful test simply be the absence of an exception

flying-sheep: done!

flying-sheep:
travis failure isn’t my fault.

flying-sheep:
@JamesHenry could you approve please?

JamesHenry:
It's in! Please rebase the PR and we'll see where we're at 😄

flying-sheep:
@JamesHenry done!

JamesHenry:
Thanks!

flying-sheep:
great!

Issue #123, opened 2016-12-08
TypeError: Cannot read property 'value' of undefined - When parsing with rule space-before-function-paren


soda0289:
**What version of TypeScript are you using?**
2.0.10

**What version of `typescript-eslint-parser` are you using?**
1.0.0

**What code were you trying to parse?**

```
export class Example {
    render () {
    }
}
```

**What did you expect to happen?**
It should pass validation

**What happened?**
Uncaught Exception
```
Cannot read property 'value' of undefined
TypeError: Cannot read property 'value' of undefined
    at EventEmitter.validateSpacingBeforeParentheses (/usr/lib/node_modules/eslint/lib/rules/space-before-function-paren.js:128:30)
    at emitOne (events.js:82:20)
    at EventEmitter.emit (events.js:169:7)
    at NodeEventGenerator.enterNode (/usr/lib/node_modules/eslint/lib/util/node-event-generator.js:40:22)
    at CodePathAnalyzer.enterNode (/usr/lib/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at CommentEventGenerator.enterNode (/usr/lib/node_modules/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter (/usr/lib/node_modules/eslint/lib/eslint.js:925:36)
    at Controller.__execute (/usr/lib/node_modules/eslint/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/usr/lib/node_modules/eslint/node_modules/estraverse/estraverse.js:501:28)
    at Controller.Traverser.controller.traverse (/usr/lib/node_modules/eslint/lib/util/traverser.js:36:33)
```

Eslint file
```
{
    "parser": "typescript-eslint-parser",
    "rules": {
        "space-before-function-paren": [
            "error",
            "always"
        ]
    }
}

```

soda0289:
Seems like babel-eslint has same problem:
https://github.com/eslint/eslint/issues/4581
https://github.com/eslint/eslint/issues/4575

monolithed:
I still see these errors:

```
/Users/a.abashkin/workspace/mail.ru/e.mail.ru/tests/node_modules/eslint/lib/rules/space-before-function-paren.js:128
            while (rightToken.value !== "(") {
                             ^

TypeError: Cannot read property 'value' of undefined
[...]

PR #124, opened 2016-12-08
Fix: Exception thrown when space occurs after function name

soda0289:
Possible fix for https://github.com/eslint/typescript-eslint-parser/issues/123

When parsing typescript node we need to add an offset when there is a space between function name and parameters.

Still need to add tests. Would like to know if I'm on the right track.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1058,7 +1058,7 @@ module.exports = function(ast, extra) {
                         generator: false,
                         expression: false,
                         body: convertChild(node.body),
-                        range: [ node.name.end, result.range[1]],
+                        range: [ node.parameters.pos - 1, result.range[1]],
                         loc: {
                             start: {
                                 line: methodLoc.line + 1,
--- a/tests/fixtures/ecma-features/classes/class-method-named-with-space.result.js
+++ b/tests/fixtures/ecma-features/classes/class-method-named-with-space.result.js
@@ -1,0 +1,320 @@
+module.exports = {
+    "type": "Program",
+    "range": [
+        0,
+        25
+    ],
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 1,
+            "column": 25
+        }
+    },
+    "body": [
+        {
+            "type": "ClassDeclaration",
+            "range": [
+                0,
+                25
+            ],
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 1,
+                    "column": 25
+                }
+            },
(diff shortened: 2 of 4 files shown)


lib/ast-converter.js
```diff
@@ -1048,6 +1048,7 @@ module.exports = function(ast, extra) {
             case SyntaxKind.GetAccessor:
             case SyntaxKind.SetAccessor:
             case SyntaxKind.MethodDeclaration:
+                var offset = node.parameters.pos - node.name.end - 1;
```
JamesHenry: This calculation ends up containing both `+node.name.end` and `-node.name.end` so we could simplify the range start to just be `node.parameters.pos - 1`, and scrap this variable altogether right?

As an aside, most of the conversion currently happens within one single function scope right now, so please give try and give variables more specific and unique names. I will be refactoring at some point once the project has fully stabilised, but just something to bear in mind at the moment.

soda0289:
@JamesHenry 
Thanks for the review. I got rid of the offset variable and just use node.parameters.pos now. I added a test and got it to pass. I noticed lots of other tests are still failing and might spend some time looking into.

JamesHenry:
Thanks a lot for your patience on this, @soda0289! As soon as https://github.com/eslint/typescript-eslint-parser/pull/131 gets merged we can see if rebasing this branch fixes the build

JamesHenry:
It's in! Please rebase the PR and we'll see where we're at 😄

soda0289:
@JamesHenry 
I reabsed off of the updated master but I can't seem to get eslintbot to be satisfied. I tried changing the last line of the commit message to fixes #123 but it still complains.

Let me know if you want anything else changed.

JamesHenry:
It is validating against our contributing guidelines.

Relevant excerpt:

> The message summary should be a one-sentence description of the change, and it must be 72 characters in length or shorter. If the pull request addresses an issue, then the issue number should be mentioned at the end. If the commit doesn’t completely fix the issue, then use (refs #1234) instead of (fixes #1234).

> Here are some good commit message summary examples:

>`Fix: Semi rule incorrectly flagging extra semicolon (fixes #840)`

Well, this change is currently breaking a lot of existing tests (causing the build to fail), so the next job will be to look into those and double-check whether or not that _should be_ the case, i.e. do the tests need updating, or does the code change? Regardless, the PR will only be merged once we get to a point where the build is passing.

soda0289:
@JamesHenry
I used plus one instead of minus one. Also one test had to be fixed. The test used spaces after function names and had a different range number.

soda0289:
Can you ensure that the range values are correct for the test I wrote and the one I changed?

JamesHenry:
LGTM, thanks!

PR #137, opened 2017-01-09
Fix async function and variable declartions

soda0289:
Async functions mark all variables as constants since the flags are not being read correctly. This patch uses the builtin typescript utilities functions to determine the variable declaration type.

Some tests are broken and I will need to add tests for this new case. Just looking for feedback.


soda0289:
Replaced with pull request #138

PR #147, opened 2017-01-24
Fix: Parameter with assignation provide type annotations (fixes #146)

### LINKED ISSUE #146: Parameters with assignations do not provide type annotation information
What version of TypeScript are you using?
2.0.3 and 2.1.5

What version of typescript-eslint-parser are you using?
1.0.2

What code were you trying to parse?
```ts
function message(name: string= "foo"): string {
    return name;
}
```

What did you expect to happen?
Parameters with assignation must also provide type annotation information.

What happened?
Parameters with assignation are not providing type annotation information.

weirdpattern:
Adds code to process the type annotations on parameters that have assignations

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1438,6 +1438,11 @@ module.exports = function(ast, extra) {
                         left: convertChild(node.name),
                         right: convertChild(node.initializer)
                     });
+                    if (node.type) {
+                        assign(result.left, {
+                            typeAnnotation: convertTypeAnnotation(node.type)
+                        });
+                    }
                 } else {
                     var convertedParameter = convert(node.name, parent);
                     if (node.type) {
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1426,29 +1426,32 @@ module.exports = function(ast, extra) {
                 break;
 
             case SyntaxKind.Parameter:
-
+                var parameter;
                 if (node.dotDotDotToken) {
+                    parameter = convertChild(node.name);
                     assign(result, {
                         type: "RestElement",
-                        argument: convertChild(node.name)
+                        argument: parameter
                     });
                 } else if (node.initializer) {
+                    parameter = convertChild(node.name);
                     assign(result, {
                         type: "AssignmentPattern",
-                        left: convertChild(node.name),
+                        left: parameter,
                         right: convertChild(node.initializer)
                     });
-                    if (node.type) {
-                        assign(result.left, {
-                            typeAnnotation: convertTypeAnnotation(node.type)
-                        });
-                    }
                 } else {
                     var convertedParameter = convert(node.name, parent);
                     if (node.type) {
                         convertedParameter.typeAnnotation = convertTypeAnnotation(node.type);
                     }
                     return convertedParameter;
+                }
+
+                if (node.type) {
+                    assign(parameter, {
+                        typeAnnotation: convertTypeAnnotation(node.type)
+                    });
                 }
 
                 break;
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks so much for your efforts, @weirdpattern! I've been so swamped in January but things are set to improve from next week.

I will try my best to get to this at the weekend, but it will be by Monday at the latest.

weirdpattern:
Perfect, could you please also take a look at issue #146, just want your input on that before I submit the PR

soda0289:
@weirdpattern I think there are some more cases that need to include type annotations

```Typescript
//Missing
function iii(...name: string): string {
        return name;
}

//Handled correctly
function jjj(name: string): string {
        return name;
}

//Fixed by this PR
function kkk(name: string = 'hello'): string {
        return name;
}

//Missing
function lll(...name: string = 'hh'): string {
        return name;
}
```

Maybe we can avoid the duplicate code by setting a variable to the argument and then setting typeAnnotation on that.

weirdpattern:
Hey @soda0289 

> ```ts 
> //Missing
> function iii(...name: Array<string>): Array<string> {
>        return name;
> }
> ```

I will look into this later today (I believe this was already being handled by the current code, but I can double check). 

> ```ts 
> //Missing
> function iii(...name: Array<string> = ['hi']): Array<string> {
>        return name;
> }
>```

I don't think this even compiles, rest operators cannot have initializers  [here](https://www.typescriptlang.org/play/#src=function%20lll(...name%3A%20Array%3Cstring%3E%20%3D%20%5B'hh'%5D)%3A%20Array%3Cstring%3E%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20name%3B%0D%0A%7D%0D%0A%0D%0Afunction%20iii(...name%3A%20Array%3Cstring%3E)%3A%20Array%3Cstring%3E%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20name%3B%0D%0A%7D)

Are you seeing something that I'm not?

soda0289:
You are right @weirdpattern those examples I gave are invalid typescript. I was just reading the code and noticied rest paremeters were missing the type annotation conversion. I did not know initializers with the rest parameter were invalid, or that it had to be an array but that makes sense.

With the following code
```typescript
function lll(...name: string[]): string {
        return name;
}

```

The output I get for function declation is

```javascript
{ type: 'FunctionDeclaration',
  range: [ 203, 267 ],
  loc: { start: { line: 16, column: 0 }, end: { line: 18, column: 1 } },
  id:
   { type: 'Identifier',
     range: [ 212, 215 ],
     name: 'lll' },
  generator: false,
  expression: false,
  async: false,
  params:
   [ { type: 'RestElement',
       range: [Object],
       loc: [Object],
       argument: [Object] } ],
  body:
   { type: 'BlockStatement',
     range: [ 243, 267 ],
     body: [ [Object] ] },
  returnType:
   { type: 'TypeAnnotation',
     range: [ 236, 242 ],
     typeAnnotation: { type: 'TSStringKeyword', range: [Object], loc: [Object] } } }

```

For the rest element
```javascript
{ type: 'RestElement',
  range: [ 216, 233 ],
  loc:
   { start: { line: 16, column: 13 },
     end: { line: 16, column: 30 } },
  argument:
   { type: 'Identifier',
     range: [ 219, 223 ],
     name: 'name' } }

```

I think type annotations are missing from the argument property of the RestElement node and would also need to have type annotations included.

weirdpattern:
@soda0289  I think you are right. I will fix this today. Thanks!

weirdpattern:
@soda0289 ok, I've included rest arguments in this PR

PR #155, opened 2017-02-07
Fix: Calculate range correctly when generic class is exported (fixes #152)

### LINKED ISSUE #152: Class body has invalid range that breaks brace-style rule
I noticed on the newer version of eslint (3.14) that the brace style rule fails on exported classes. The opening brace variable used in the rule is incorrect since it is pointing to the identifier 'hello'. I believe the class body node's range is wrong and is still including the identifier. Have not tested on typescript 2.0.3. Will try and submit pr with fix.

2.1.3

ts-2.1

```
export class hello {

    hello () {

    }
}

```

Should parse correctly with brace-style 1tbs

**What happened?**
Failed with error:
  2:14  error  Statement inside of curly braces should be on next line  brace-style

soda0289:
When a class with a generic definition is exported we must take the
type parameters into account when finding the opening brace. This
continues the fix of pull request #153 and handles the case of exporting
a class with generics. We now find the last type parameter and use the
succeeding token as the opening brace.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1465,6 +1465,13 @@ module.exports = function(ast, extra) {
                 var lastClassToken = heritageClauses.length ? heritageClauses[heritageClauses.length - 1] : node.name;
                 var classNodeType = SyntaxKind[node.kind];
 
+                if (node.typeParameters && node.typeParameters.length) {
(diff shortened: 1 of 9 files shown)


soda0289:
I missed a case earlier. I have tested this with extends and implements clauses as well. Let me know if I should add tests for theses cases. We already have tests for implements and extends but not with export.

PR #163, opened 2017-02-19
Fix: Label Functions and Methods declartions as Ambient (fixes #162)

### LINKED ISSUE #162: Declared class methods fail rules that depend on function body
Eslint: 3.15 (with eslint-scope#master)
TypeScript: next
typescipt-eslint-parser: master

```ts
declare namespace FF {
    class Foo extends Bar.Baz {
        far(): any;
    }
}
```
or
```ts
declare module "FF" {
    class Foo extends Bar.Baz {
        far(): any;
    }
}
```
or
```ts

declare class Foo extends Bar.Baz {
    far(): any;
}

```

It should parse

**What happened?**
Exception:
```
Cannot read property 'type' of null
TypeError: Cannot read property 'type' of null
    at EventEmitter.onCodePathStart (/home/reyad/Workspace/eslint/lib/rules/array-callback-return.js:187:34)
    at emitTwo (events.js:92:20)
    at EventEmitter.emit (events.js:172:7)
    at processCodePathToEnter (/home/reyad/Workspace/eslint/lib/code-path-analysis/code-path-analyzer.js:357:30)
    at CodePathAnalyzer.enterNode (/home/reyad/Workspace/eslint/lib/code-path-analysis/code-path-analyzer.js:604:9)
    at CommentEventGenerator.enterNode (/home/reyad/Workspace/eslint/lib/util/comment-event-generator.js:98:23)
    at Controller.enter (/home/reyad/Workspace/eslint/lib/eslint.js:928:36)
    at Controller.__execute (/home/reyad/Workspace/eslint/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/home/reyad/Workspace/eslint/node_modules/estraverse/estraverse.js:501:28)
    at Controller.Traverser.controller.traverse (/home/reyad/Workspace/eslint/lib/util/traverser.js:36:33)
```

We should detect if the parent is being declared and not use FunctionExpression to define functions that have no body. Similar to what is done with functions in a namespace.

soda0289:
Ambient functions do not have a body and will cuase rules to throw
an exception. We use the types TSAmbientFunctionExpression and
TSAmbientMethodDefinition.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -875,7 +875,7 @@ module.exports = function(ast, extra) {
                         return modifier.kind === ts.SyntaxKind.DeclareKeyword;
                     });
                     if (isDeclareFunction) {
-                        functionDeclarationType = "DeclareFunction";
+                        functionDeclarationType = "TSAmbientFunctionDeclaration";
                     }
                 }
 
@@ -1074,6 +1074,7 @@ module.exports = function(ast, extra) {
                 // TODO: double-check that these positions are correct
                 var methodLoc = ast.getLineAndCharacterOfPosition(node.name.end + 1),
                     nodeIsMethod = (node.kind === SyntaxKind.MethodDeclaration),
+                    isAmbient = ts.isInAmbientContext(node),
                     method = {
                         type: "FunctionExpression",
                         id: null,
@@ -1136,6 +1137,10 @@ module.exports = function(ast, extra) {
                             methodDefinitionType = "TSAbstractMethodDefinition";
                         }
                     }
+                    if (isAmbient) {
+                        methodDefinitionType = "TSAmbientMethodDefinition";
+                        method.type = "TSAmbientFunctionExpression";
+                    }
 
                     assign(result, {
                         type: methodDefinitionType,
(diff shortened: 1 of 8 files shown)


soda0289:
I noticed in FunctionDeclation we use the type DeclareFunction when defining ambient function. Should we use TSAmbientFunctionDeclation as well?

soda0289:
I made the change for the node DeclareFunction is now TSAmbientFunctionDeclaration to make it align with this pull request. At the very least we should prefix it with TS unless there is another rule that uses DeclareFunction.

Pajn:
When I looked at the test output of this `declare class` test for #207 I noticed that there is nothing on the class node that make it possible to see that it is declared. That info would be great for prettier so that it can now if it should print declare or not :)
It can of course be added in a different PR, just saw it here.

soda0289:
@Pajn Yes we do need to label ambient classes. I was thinking of changing the type from Class to TSAmbientClass this way some eslint rules won't crash as they are not aware of empty body methods, which declared classes all have. This would break current linting rules that are meant for classes but it is better then having the crash.

JamesHenry:
I'm so sorry it's taken so long to address these PRs @soda0289, so happy to have you on the team now! Let's take stock as soon as we merge in all the breaking changes and release v3!

PR #168, opened 2017-02-22
Fix: Missing parameter properties info in constructors (fixes #143)

### LINKED ISSUE #143: Parameter properties information not being capture in ESTree parameter node
2.0.10 and 2.1.5

1.0.2

```ts
class Foo {
    constructor(public name : string) {}
}
```

```ts
class Foo {
    constructor(protected name : string) {}
}
```

```ts
class Foo {
    constructor(private name : string) {}
}
```

```ts
class Foo {
    constructor(readonly name : string) {}
}
```

```ts
class Foo {
    constructor(public readonly name : string) {}
}
```

```ts
class Foo {
    constructor(protected readonly name : string) {}
}
```

```ts
class Foo {
    constructor(private readonly name : string) {}
}
```

Parameter properties should be parsed correctly and modifiers should be part of the final ESTree parameter node.

**What happened?**
The modifiers property of the TS Node is being ignored by the AST Converter. The final ESTree contains the parameters, but without the accessibility modifiers and/or readonly modifiers.

Sources: [Parameter properties](http://www.typescriptlang.org/docs/handbook/classes.html#parameter-properties)

weirdpattern:
Includes modifications to ast-converter and corresponding tests

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1187,10 +1187,27 @@ module.exports = function(ast, extra) {
                         id: null,
                         params: node.parameters.map(function(param) {
                             var convertedParam = convertChild(param);
-                            convertedParam.decorators = (param.decorators) ? param.decorators.map(function(d) {
+                            var decorators = (param.decorators) ? param.decorators.map(function(d) {
                                 return convertChild(d.expression);
                             }) : [];
-                            return convertedParam;
+
+                            if (param.modifiers) {
+                                return {
+                                    type: "TSParameterProperty",
+                                    range: [param.getStart(), param.end],
+                                    loc: getLoc(param, ast),
+                                    accessibility: getTSNodeAccessibility(param),
+                                    isReadonly: param.modifiers.some(function(modifier) {
+                                        return modifier.kind === SyntaxKind.ReadonlyKeyword;
+                                    }),
+                                    parameter: convertedParam,
+                                    decorators: decorators
+                                };
+                            }
+
+                            return assign(convertedParam, {
+                                decorators: decorators
+                            });
                         }),
                         generator: false,
                         expression: false,
--- a/tests/fixtures/typescript/basics/class-with-private-parameter-properties.result.js
+++ b/tests/fixtures/typescript/basics/class-with-private-parameter-properties.result.js
@@ -1,0 +1,1131 @@
+module.exports = {
+    "type": "Program",
+    "range": [
+        0,
+        203
+    ],
+    "loc": {
+        "end": {
+            "line": 6,
+            "column": 1
+        },
+        "start": {
+            "line": 1,
+            "column": 0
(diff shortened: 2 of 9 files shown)


weirdpattern:
Build is not being updated to complete, although travis completed successfully...

soda0289:
@weirdpattern Maybe a webhook failed from traivs. No worries.

LGTM, Thanks!

soda0289:
@JamesHenry Could you take a look at this one too. It would allow for us to implement the TSLint rule no-parameter-properties for the typescript plugin.

JamesHenry:
@weirdpattern @soda0289 Sorry to be a pain, but could we do an initial PR which adds a test for a standard JS class with one or more parameters in its constructor?

I was surprised to see that we don't already have one in our ecmaFeatures/classes fixtures.

If we could get that merged in first, it would be a really useful reference within this PR to see a before and after. It would also make it much easier to identify and unintended issues.

JamesHenry:
There may well be no change

weirdpattern:
sure, no problem... I can work on this tonight unless @soda0289 wants to do it now, I leave the decision to you 😀

soda0289:
I can write some tests, I don't mind

weirdpattern:
@soda0289 go for it and thank you!

JamesHenry:
Ready to be rebased against master! 😄

soda0289:
@weirdpattern I think you did a merge instead of a rebase and force push

weirdpattern:
@soda0289 I think you are right, I was sure I did a rebase, no idea what I did wrong ... what should I do now?

soda0289:
@weirdpattern Perfrom the rebase again it should clean it up. When you perfrom the push it might warn you that the histroy has changed. Just ignore the message and do a git push --force

weirdpattern:
@soda0289 ok I think I got it right this time... please review

soda0289:
@weirdpattern Yup! LGTM, Thanks!

lib/ast-converter.js
```diff
@@ -1187,10 +1187,27 @@ module.exports = function(ast, extra) {
...
                             }) : [];
-                            return convertedParam;
+
+                            if (param.modifiers) {
+                                return {
+                                    type: "TSParameterProperty",
+                                    range: [param.getStart(), param.end],
+                                    loc: getLoc(param, ast),
+                                    accessibility: getTSNodeAccessibility(param),
+                                    isReadonly: param.modifiers.filter(function(modifier) {
```
JamesHenry: Using filter is a bit of an odd way to check this. More semantic would be:

```
isReadonly: param.modifiers.some(function(modifier) {
  return modifier.kind === SyntaxKind.ReadonlyKeyword
})
```

weirdpattern:
@JamesHenry done

PR #176, opened 2017-02-26
Fix: Add start and end propert to tokens (fixes #172)

### LINKED ISSUE #172: getting `Cannot read property 'type' of undefined` on tsx file
2.2.1

2.0.0

```

class App extends React.PureComponent<void, void> {
  render () {
    return (
      <Provider store={store}>
        <View style={{ flex: 1 }}>
          <StatusBar barStyle='light-content' />
          <Navigator />
          <Loader />
          <AlertModal />
        </View>
      </Provider>
    );
  }
}

```

**What happened?**

I get warning `Cannot read property 'type' of undefined` from vscode
Error from console
```
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at EventEmitter.Literal (/Users/LLu/Code/Patient/client/node_modules/eslint/lib/rules/no-empty-character-class.js:49:26)
```

**Extra info**
I'm using eslint plugin with vscode

**eslint.rc**
```
{
  "parser": "typescript-eslint-parser",
  "parserOptions": {
    "ecmaVersion": 6,
    "sourceType": "module",
    "ecmaFeatures": {
        "jsx": true
    }
  },
  "rules": {
    "no-empty-character-class":"error"
  }
} 
```

soda0289:
Some eslint plugins except tokens to have start and end properties
instead of range. This commit adds those properties and modifies
the tests to ignore start and end property.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -410,6 +410,8 @@ function convertToken(token, ast) {
         newToken = {
             type: getTokenType(token),
             value: value,
+            start: start,
+            end: token.end,
             range: [start, token.end],
             loc: getLoc(token, ast)
         };
--- a/tests/lib/attach-comments.js
+++ b/tests/lib/attach-comments.js
@@ -34,7 +34,8 @@ var assert = require("chai").assert,
     leche = require("leche"),
     path = require("path"),
     parser = require("../../parser"),
-    shelljs = require("shelljs");
+    shelljs = require("shelljs"),
+    tester = require("./tester");
 
 //------------------------------------------------------------------------------
 // Setup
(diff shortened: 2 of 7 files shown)


JamesHenry:
What is the motivation behind hiding them from the tests?

soda0289:
Well it would be a lot of work to rewrite every test. I have seen the tool we have that might help but there are still JSON tests that would need to have start and end inserted in every token node.

I looked at how espree handles this and copied the code from them.

Another way to not include them would be to add the property as non-enumerable and that way when we run JSON.stringify start and end property would not show up. 

I could write a tool that could update the JSON tests and upadte the tool we have to update all the tests, not just the typescript ones. 

Let me know if you think its worth it or if you think making the property non-enumerable is cleaner

JamesHenry:
Ah yes, I had forgotten about that! I think using espree's example is more than ok 😄 

Thanks so much for providing PRs for the issues you find, it's so helpful!

Issue #118, opened 2016-11-24
Parser does not handle jsdoc well


jtheoof:
**What version of TypeScript are you using?**
2.1.1

**What version of `typescript-eslint-parser` are you using?**
1.0.0

**What code were you trying to parse?**

```
/**
 * This crashes eslint.
 */
export interface Test {
    name: string;
}
```

**What did you expect to happen?**
Syntax should parse correctly

**What happened?**
Stacktrace:

```
Cannot read property 'loc' of undefined
TypeError: Cannot read property 'loc' of undefined
    at getNodeIndent (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/rules/indent.js:236:54)
    at EventEmitter.Program (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/rules/indent.js:793:49)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
    at NodeEventGenerator.enterNode (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/util/node-event-generator.js:40:22)
    at CodePathAnalyzer.enterNode (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
    at CommentEventGenerator.enterNode (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/util/comment-event-generator.js:97:23)
    at Controller.enter (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/eslint/lib/eslint.js:903:36)
    at Controller.__execute (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/estraverse/estraverse.js:397:31)
    at Controller.traverse (/Users/jattali/dev/mnubo/mnubo-front-end/app/node_modules/estraverse/estraverse.js:501:28)
```

Note: If you remove the `jsdoc` comments, `eslint` does not crash anymore.

jtheoof:
In fact, if you replace:

```
/**
 * This crashes eslint.
 */
```

with:

```
/*
 * This crashes eslint.
 */
```

`eslint` is happy again.

jtheoof:
Probably related to #5

weirdpattern:
[...]

PR #177, opened 2017-02-26
Fix: Remove jsdoc node property from ts nodes (fixes #164)

### LINKED ISSUE #164: JSDoc in interface with @ symbol causes node.getStart() to throw exception
TypeScript: master
typescript-eslint-parser: ts-2.1
Eslint: 3.15

```ts
interface Test {
    /**
     * Comment Line 1
     * @whatever bar 
     */
    foo(bar);
}
```

It should parse

**What happened?**
Exception from typescript when calling node.start();
```
TypeError: Cannot read property 'text' of undefined
    at Object.getTokenPosOfNode (/home/reyad/Workspace/typescript-eslint-parser/node_modules/typescript/lib/typescript.js:6807:71)
    at TokenObject.TokenOrIdentifierObject.getStart (/home/reyad/Workspace/typescript-eslint-parser/node_modules/typescript/lib/typescript.js:81993:23)
    at convert (/home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:490:30)
    at convertChild (/home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:517:20)
    at /home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:699:39
    at Array.forEach (native)
    at deeplyCopy (/home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:692:16)
    at convert (/home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:2043:17)
    at convertChild (/home/reyad/Workspace/typescript-eslint-parser/lib/ast-converter.js:517:20)
    at Array.map (native)

```

soda0289:
JSDoc property is not found on parsed ESTree nodes and should be
removed when parsing typescript nodes.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -669,7 +669,7 @@ module.exports = function(ast, extra) {
         function deeplyCopy() {
             result.type = "TS" + SyntaxKind[node.kind];
             Object.keys(node).filter(function(key) {
-                return !(/^(?:kind|parent|pos|end|flags|modifierFlagsCache)$/.test(key));
+                return !(/^(?:kind|parent|pos|end|flags|modifierFlagsCache|jsDoc)$/.test(key));
             }).forEach(function(key) {
                 if (key === "type") {
                     result.typeAnnotation = (node.type) ? convertTypeAnnotation(node.type) : null;
--- a/tests/fixtures/typescript/basics/interface-with-jsdoc.result.js
+++ b/tests/fixtures/typescript/basics/interface-with-jsdoc.result.js
@@ -1,0 +1,654 @@
+module.exports = {
+    "type": "Program",
+    "range": [
+        0,
+        87
+    ],
+    "loc": {
+        "start": {
+            "line": 1,
+            "column": 0
+        },
+        "end": {
+            "line": 7,
+            "column": 1
+        }
+    },
+    "body": [
+        {
+            "type": "TSInterfaceDeclaration",
+            "range": [
+                0,
+                87
+            ],
+            "loc": {
+                "start": {
+                    "line": 1,
+                    "column": 0
+                },
+                "end": {
+                    "line": 7,
+                    "column": 1
+                }
+            },
+            "name": {
+                "type": "Identifier",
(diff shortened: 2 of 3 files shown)


soda0289:
I looked at parsed espree and acorn nodes, they do not include the jsdoc property so I dont think we would need them on ts nodes since we dont have them on estree nodes.

soda0289:
I added a test

JamesHenry:
This is currently not doing any harm right?

We have started planning ESLint v4 which will contain the comment attribution work, so I think I would prefer to wait and address the comments situation holistically.

Please let me know if this is actually an issue

soda0289:
When we convert Class, Function, or MethodDeclaration nodes we strip out the jsDoc comment property.
We do not do this on typescript nodes like Interface. The example code I posted in issue #164 will not parse and causes the parser to throw an exception. The exception is caused by the JSDocTag failing on node.getStart(). This used to work in older version of typescript I think. 

I do not see any harm in remove this property since it is not included on other parsers like acorn or espree. We can wait if you like but I think having this parser not fail or throw uncaught exceptions is really important.

soda0289:
I could look into the typescript code base or post an issue to figure out why node.getStart() is failing?

weirdpattern:
take a look at #118, this was closed a while, but refers to the exact same issue @soda0289 is talking about...

soda0289:
@weirdpattern Issue #118 was a bug in typescript where it would fail to produce tokens correctly for JSDoc comments. This issue is a little different as it deals with converting typescript JSDoc nodes into non existing ESTree nodes.

weirdpattern:
by the way, node.getStart() fails because the node does not declare a parent... I looked into this when #118 was still open =]

soda0289:
@weirdpattern I noticed that the parent property was null which means it probably is a bug in typescript, since it did work in typescript 2.0.

JamesHenry:
Ah sorry @soda0289 I did misunderstand 😄 ! Yes let's get this in

PR #178, opened 2017-02-26
Chore: Add test for constructor with parameters

soda0289:
Here is a test that has a class constructor with two parameters. 

I noticed we are missing a lot of tests for newer version of ecmascript like async/await functions. As well as tests for class methods with default parameters, deconstructed parameters, and rest parameters. 

@JamesHenry Let me know if you want me to add more tests.

### DIFF
--- a/tests/fixtures/ecma-features/classes/class-with-constructor-parameters.result.js
+++ b/tests/fixtures/ecma-features/classes/class-with-constructor-parameters.result.js
@@ -1,0 +1,413 @@
+module.exports = {
+    "type": "Program",
+    "range": [
+        0,
+        33
+    ],
+    "loc": {
+        "start": {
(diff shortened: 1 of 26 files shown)


weirdpattern:
@JamesHenry do we need a test with initializers and rest params as well?

weirdpattern:
the current test LGTM

JamesHenry:
Yes please! That would be perfect. Looks great so far @soda0289

soda0289:
@weirdpattern @JamesHenry I have added tests for rest parameters, default parameters, and destructing parameters for both class method and constructors.

Let me know if you think we need any more

weirdpattern:
LGTM, thank you @soda0289

JamesHenry:
Great, thanks!

PR #181, opened 2017-03-06
Fix: Change DoWhileStatement to DoStatement to match TS (fixes #180)

### LINKED ISSUE #180: TypeScript calls DoWhileStatement a DoStatement
This means that the current AST we are outputting for DoWhileStatements is incorrect (we are auto-converting the node as TSDoStatement)

JamesHenry:


JamesHenry:
Thanks, no they don't! This was just an untested typo I guess. DoWhileStatements are not exactly that common in JS so I think that it why it was missed for so long

PR #183, opened 2017-03-17
New: Add type arguments and parameters to AST

Pajn:
For prettier/prettier/issues/13 this information must be preserved in the AST.
I had no idea how to create the test so I just did what seemed reasonable but please correct me if it's wrong.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -546,7 +546,7 @@ module.exports = function(ast, extra) {
                             typeArgument.end
                         ],
                         loc: getLocFor(typeArgumentStart, typeArgument.end, ast),
-                        id: convertChild(typeArgument.typeName)
+                        id: convertChild(typeArgument.typeName || typeArgument)
                     };
                 })
             };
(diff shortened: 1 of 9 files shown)


soda0289:
I think you will want to use `convertTSTypeParametersToTypeParametersDeclaration()` to convert the typescript type parameters instead of `convertTypeArgumentsToTypeParameters()`

I'm not a 100% sure thou. @JamesHenry do you why there are two functions it seems one is only used for class declarations?

The tests look fine to me.

Pajn:
I understood it as `convertTSTypeParametersToTypeParametersDeclaration` is for declaring them (introducing them in the scope) and `convertTypeArgumentsToTypeParameters` is for passing them, which is what is done in CallExpressions.

I realized that primitive types did not work, which is in my second commit. I'm unsure if I should include that in this PR or not. But I guess the above confusion must be solved first to even see if I modified the correct function :)

soda0289:
@Pajn you could be correct. I just saw that `convertTypeArgumentsToTypeParameters()` is only used for class declarations.

Reading the code I don't think the new expression `const a = new A<B>();` is also not keeping the type arguments. Is this correct? Do you want to add support for this as well?

soda0289:
I also think class declarations do not record the generic type argument on the implement clauses do. For example:
``ts
class A<T> {

}
``

Pajn:
Good catches! Might as well get ass many cases as possible fixed.

soda0289:
Nice looks good!

JamesHenry:
This LGTM, thanks @Pajn! It just needs to be rebased against master

Pajn:
Done. I was unsure on New/Fix again, I can change to what you prefer.

I must say thank you to both of you for being both helpful and very responsive on these PRs!

PR #187, opened 2017-03-17
New: Mark optional parameters (fixes #186)

### LINKED ISSUE #186: Whenever a parameter is optional ot not should be stored in the AST
e in this example should have an `optional: true` flag
```typescript
function processEntity(e?: Entity) {
    validateEntity(e);
    let s = e!.name;
}
```

Pajn:
I choose to avoid storing `optional: false` to avoid a giant mess of failing tests, as undefined is false the use case still works.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -752,6 +752,14 @@ module.exports = function(ast, extra) {
                     type: "Identifier",
                     name: ts.unescapeIdentifier(node.text)
                 });
+                if (node.parent.questionToken && (
+                    SyntaxKind.Parameter === node.parent.kind ||
+                    SyntaxKind.PropertyDeclaration === node.parent.kind ||
+                    SyntaxKind.PropertySignature === node.parent.kind ||
+                    SyntaxKind.MethodSignature === node.parent.kind
+                )) {
+                    result.optional = true;
+                }
                 break;
 
             case SyntaxKind.WithStatement:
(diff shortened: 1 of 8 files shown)


Pajn:
We would need everything to be able to print the AST back to source. This was just what I encountered but to avoid a bunch of PRs to extend that if I can try to lookup and fix other cases. Apart from those two, properties in type object literals come to mind as well but I guess there can be even more.

Pajn:
I think I have gotten all cases, however the SyntaxKinds does not align perfectly with the spec so if you think of any other case to test please alert me.
Type aliases initializers goes trough the same path as type annotations so I didn't add en explicit test for those but can of course do so if you prefer.

JamesHenry:
I think this is good for now. I will likely refactor this a bit in a future PR to have consistency around the `optional`property (i.e. add the `optional: false` to other tests)

JamesHenry:
Blast, this was out of date with master and has caused an issue with the new optional params. I will submit a fix. I thought github repo was configure to block situations like that

JamesHenry:
Purely a test output that needs updating, no functional regression

Pajn:
I can rebase the PRs on each other if you prefer. Just tell me which and in what order.

JamesHenry:
We'll just rebase each PR against master ahead of merging

platinumazure:
@JamesHenry Another possibility is re-trigger the Travis PR build before merging a PR if you had merged a few other PRs before getting to that one- Travis will run the PR against latest master.

JamesHenry:
Good point, @platinumazure! Thanks

PR #191, opened 2017-03-18
New: Add type annotation to class properties (fixes  #190)

### LINKED ISSUE #190: Type Annotation on ClassParameter not preserved in AST
For example in this code
```
class Foo {
  private bar : string;
  public static baz : number;
  public getBar () {
    return this.bar;
  }
  protected setBar (bar : string) {
    this.bar = bar;
  }
}
```
the `string` and `number` information is lost

Pajn:
Type information is preserved for interface properties and object type literals, seems to just be missing on class properties.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1082,7 +1082,8 @@ module.exports = function(ast, extra) {
                     accessibility: getTSNodeAccessibility(node),
                     decorators: (node.decorators) ? node.decorators.map(function(d) {
                         return convertChild(d.expression);
-                    }) : []
+                    }) : [],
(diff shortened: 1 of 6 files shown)


JamesHenry:
Hi, @Pajn! Thanks a lot for contributing (and starting the PR for more prettier work). I am excited about what that can become.

This LGTM, I think it should really be a `Fix: ` commit, though. I will amend it whilst merging in this case.

Pajn:
>I think it should really be a Fix: commit, though. I will amend it whilst merging in this case.
Ah, sorry. Still learning your conventions :)

Your integration of this project into prettier made it easy to incrementally improve TS support so huge thanks for that!

JamesHenry:
For sure! Let's make this happen 😄

PR #192, opened 2017-03-18
Fix: Updated broken class-with-optional-properties test result

JamesHenry:


platinumazure:
Waiting for Travis to complete, then I'll approve :smile:

PR #193, opened 2017-03-18
Update: Open TS peerDependency, warn non-supported version (fixes #167)

### LINKED ISSUE #167: typescript peerDependency blocks npm shrinkwrap
Please use `"typescript": "*"` or at least `"typescript": "^2.0.0"` in `peerDependencies` as current `"typescript": "~2.0.3"` means that typescript `<2.1.0` is required.

This happens on my machine with `"typescript": "2.1.6",` and `"typescript-eslint-parser": "1.0.3",` in package.json:
```
npm WARN typescript-eslint-parser@1.0.3 requires a peer of typescript@~2.0.3 but none was installed.
npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\Zaggi\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "shrinkwrap"
npm ERR! node v7.5.0
npm ERR! npm  v4.2.0

npm ERR! Problems were encountered
npm ERR! Please correct and try again.
npm ERR! peer invalid: typescript@~2.0.3, required by typescript-eslint-parser@1.0.3
```

JamesHenry:
This would be one way to address https://github.com/eslint/typescript-eslint-parser/issues/167

Still not sure how I feel about it, but thought I would stick a working example on a branch and look for feedback.

@soda0289 @eslint/eslint-team

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,9 +52,10 @@
   },
   "dependencies": {
     "lodash.unescape": "4.0.0",
-    "object-assign": "^4.0.1"
+    "object-assign": "^4.0.1",
+    "semver": "^4.3.6"
   },
   "peerDependencies": {
-    "typescript": "~2.2.1"
+    "typescript": "*"
   }
 }
--- a/README.md
+++ b/README.md
@@ -6,9 +6,13 @@ A parser that converts TypeScript into an [ESTree](https://github.com/estree/est
 
 ## Supported TypeScript Version
 
-The version of TypeScript supported by this parser is `~2.2.1`. This is reflected in the `peerDependency` requirement within the package.json file.
+We will always endeavor to support the latest stable version of TypeScript.
 
-**Please ensure that you are using this version before submitting any issues.**
+The version of TypeScript currently supported by this parser is `~2.2.1`. This is reflected in the `devDependency` requirement within the package.json file, and it is what the tests will be run against. We have an open `peerDependency` requirement in order to allow for experimentation on newer/beta versions of TypeScript.
+
+If you use a non-supported version of TypeScript, the parser will log a warning to the console.
+
+**Please ensure that you are using a supported version before submitting any issues/bug reports.**
 
 ## Usage
 
--- a/parser.js
+++ b/parser.js
@@ -9,7 +9,27 @@
 "use strict";
 
 var astNodeTypes = require("./lib/ast-node-types"),
-    ts = require("typescript");
+    ts = require("typescript"),
+    semver = require("semver");
+
+var SUPPORTED_TYPESCRIPT_VERSIONS = require("./package.json").devDependencies.typescript;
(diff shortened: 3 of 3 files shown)


soda0289:
I like this solution. It gives the user more information on what to expect.

For example typescript 2.3 has some breaking changes again. This time with how JSX is handled, need to spend some more time looking into. When typescript upgrades we will have to release a new version and a warning will help user understand why the parser is failing.

corbinu:
This feels like the best solution to me also

alberto:
Not sure that will work for non-final releases: http://blog.npmjs.org/post/115305091285/introducing-the-npm-semantic-version-calculator

soda0289:
We currently use `~2.2.1` for the typescript devDependency. Which should only include versions between >=2.2.0 and < 2.3.0. According to the calculator, https://semver.npmjs.com/, it will only include version 2.2.1 and not the insider preview 2.2.2-insiders.20170302  or any of the development releases like 2.3.0-dev.20170318. I can't think of any other cases where this would fail.

soda0289:
@JamesHenry 
I think we may need better logic to get the globally installed version of typescript. If we `require('typescript')` that will only check the devDependency version, which would always match the package.json, not the globally installed one.

soda0289:
The more I think about this I think it will be fine since devDepencies are not installed. It should work as expected.

JamesHenry:
@soda0289 The reason why we need it as a peerDependency in the first place is that we `require('typescript')` and use it as a library. A global installation of TypeScript gives you access to the compiler's CLI, but that is not our use-case. The require will use whatever version the user installed in their project, the devDependency is only relevant for this project when it runs its tests.

soda0289:
That makes sense. I was thinking that typescript was bundled with this parser but that is not the case and only happens during development.

JamesHenry:
Ok I am going to clean this up and request review

package.json
```diff
@@ -52,9 +52,10 @@
   },
   "dependencies": {
     "lodash.unescape": "4.0.0",
-    "object-assign": "^4.0.1"
+    "object-assign": "^4.0.1",
+    "semver": "^4.3.6"
```
SimenB: any particular reason this is not using 5?

soda0289: Can't think of one. I looked at the commit log and seems they only dropped support for AMD modules and minifying the code in v5. Do you want to create a PR?

SimenB: Sure!

Interesting message on running `npm install` here, btw.

`npm WARN The package semver is included as both a dev and production dependency.`

soda0289: We don't need it in `devDependencies` anymore as it is used during runtime.

SimenB: #246

PR #195, opened 2017-03-18
Fix: Convert MetaProperty (new.target) nodes correcly (fixes #194)

### LINKED ISSUE #194: MetaProperty nodes are not converted correctly
2.2.1

2.0.0

```ts
new foo.bar();
//OR
() => {
    new.target;
}
```

`new foo.bar()` should generate a NewExpression node
`new.target()` should generate a MetaProperty node

**What happened?**
They both generate invalid MetaProperty nodes

PR will be up soon

soda0289:
The MetaProperty node was added a couple of months ago into typescript. I think it landed in 2.2.
https://github.com/Microsoft/TypeScript/pull/12783

I have to create a identifier node for 'new' since it is not part of the typescript metaproperty node.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1819,23 +1819,27 @@ module.exports = function(ast, extra) {
                 break;
 
             case SyntaxKind.NewExpression:
-
-                // new.target
-                if (node.expression.kind === SyntaxKind.PropertyAccessExpression) {
-                    assign(result, {
-                        type: "MetaProperty",
-                        meta: convertChild(node.expression.name),
-                        property: convertChild(node.expression.propertyName)
-                    });
-                } else {
-                    assign(result, {
-                        type: "NewExpression",
-                        callee: convertChild(node.expression),
-                        arguments: (node.arguments) ? node.arguments.map(convertChild) : []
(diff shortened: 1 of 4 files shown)


tests/lib/basics.js
```diff
@@ -59,6 +59,9 @@ describe("basics", function() {
             try {
                 result = parser.parse(code, config);
                 result = tester.getRaw(result);
+                if (Object.keys(expected).length === 0) {
+                    console.log(JSON.stringify(result, null, 4));
+                }
```
soda0289: I got to delete this. Sorry

JamesHenry: Just in time 😄  I was just reviewing

soda0289:
We have those tests already:
https://github.com/eslint/typescript-eslint-parser/tree/master/tests/fixtures/ecma-features/newTarget

I had to enable the new-target rule by removing it from the filter in tests/lib/ecma-features.js

Typescript does not throw an error you have an invalid metaproperty, such as `new.unkown_property`. I could add tests for those cases.

JamesHenry:
Ah! Sorry, that makes sense

PR #196, opened 2017-03-19
Breaking: Normalzie type parameters (fixes #197)

### LINKED ISSUE #197: Normalize type arguments and parameters for all nodes
To make the output more consistent `deeplyClone` should normalize type arguments and type parameters as well.

Pajn:
The `deeplyCopy` function does a great job of supporting strange TS nodes however I run in to a few where it would have been nice if it had normalized type arguments and type parameters just as for "supported" nodes.
The nodes in question are
For type arguments:
- TSConstructorType
- TSFunctionType
- TSTypeReference

For type parameters:
- TSCallSignature
- TSConstructSignature

Now this do break some tests because previously such nodes were outputted as they appeared in the TypeScript AST and is therefore a breaking change. If this actually breaks anyone though is harder to tell, as far as I understand there aren't many TS specific Eslint rules around which could have been using these props. I have not updated any tests yet so you can see the breakage in the Travis logs.

Alternatives for doing this:
1. Specifically support these nodes.
  TypeScript have a lot of strange syntax so it's a risk that there are more types than the above listed. Specifically supporting these nodes would limit the impact on today unknown types. This would however likely plan out in me opening PRs to support these unknown nodes when I found them.
2. Not doing this at all.
  The `deeplyCopy` function already outputs enough to support these nodes without any change in `typescript-eslint-parser` however that would require prettier to support both formats as supported nodes already are normalized.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -709,6 +709,14 @@ module.exports = function(ast, extra) {
             }).forEach(function(key) {
                 if (key === "type") {
                     result.typeAnnotation = (node.type) ? convertTypeAnnotation(node.type) : null;
+                } else if (key === "typeArguments") {
+                    result.typeParameters = (node.typeArguments)
+                        ? convertTypeArgumentsToTypeParameters(node.typeArguments)
+                        : null;
+                } else if (key === "typeParameters") {
+                    result.typeParameters = (node.typeParameters)
+                        ? convertTSTypeParametersToTypeParametersDeclaration(node.typeParameters)
+                        : null;
                 } else {
                     if (Array.isArray(node[key])) {
                         result[key] = node[key].map(convertChild);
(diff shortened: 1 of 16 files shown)


JamesHenry:
The only known downstream consumer is https://github.com/nzakas/eslint-plugin-typescript. I am onboard with this change, I think we would have done it in the first place, but the `deeplyCopy` was added long before the other helpers.

Pajn:
Great! I will fix some tests then

Pajn:
I updated the commit message. Is there anything else I should do as well?

soda0289:
@Pajn Looks like the tests are failing. I'm pretty sure they were working before. Did you forget to push some changes?

Pajn:
Hmm, maybe some other commit intervene. I'll rebase and fix them ASAP, might be next week though.

JamesHenry:
Hi @Pajn, we have another breaking change to merge in here https://github.com/eslint/typescript-eslint-parser/pull/218, and so it would be great to get them both in together.

When do you think you will have chance to fix the commit?

JamesHenry:
And another https://github.com/eslint/typescript-eslint-parser/pull/233 :)

Pajn:
Sorry for not responding. I have been very busy with a move to a different city.
I've now rebased it.

JamesHenry:
Thanks, @Pajn!

PR #199, opened 2017-03-19
Add note about babylon AST

ChristopherBiscardi:
When trying to add nodes converting from typescript `ts.SyntaxKind.EnumDeclaration`s to "ESTree" I was looking at the [`VariableDeclaration`/`VariableDeclarator`](https://github.com/eslint/typescript-eslint-parser/blob/bfb1506c48b625871ffeb67dbec7941d460f8941/lib/ast-converter.js#L944-L965) conversions and it took me a bit to realize they were [babylon AST](https://github.com/babel/babylon/blob/master/ast/spec.md#variabledeclaration) instead of [ESTree](https://github.com/estree/estree/blob/master/es2015.md#variabledeclaration).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # TypeScript ESLint Parser (Experimental)
 
-A parser that converts TypeScript into an [ESTree](https://github.com/estree/estree)-compatible form so it can be used in ESLint. The goal is to allow TypeScript files to be parsed by ESLint (though not necessarily pass all ESLint rules).
+A parser that converts TypeScript into an [ESTree](https://github.com/estree/estree)-compatible form (specifically, the [babylon AST](https://github.com/babel/babylon/blob/master/ast/spec.md)) so it can be used in ESLint. The goal is to allow TypeScript files to be parsed by ESLint (though not necessarily pass all ESLint rules).
 
 **Important:** This parser is still in the very early stages and is considered experimental. There are likely a lot of bugs. You should not rely on this in a production environment yet.
 


soda0289:
I'm not sure if the goal of the project is to convert typescript AST into babylon AST. I do agree we need better documentation about what we output and the extension we make on the ESTree nodes. Typescript has a lot of unique features so it is not possible to use the bablyon ast for everything but it is a good starting point.

nzakas:
Sorry, I'm having a hard time understanding what you're saying. We aren't using Babylon AST. Can you explain more about why you think we are?

ChristopherBiscardi:
@nzakas of course, sorry for being unclear. I was looking at [this code in ast-converter](https://github.com/eslint/typescript-eslint-parser/blob/bfb1506c48b625871ffeb67dbec7941d460f8941/lib/ast-converter.js#L944-L965) which uses the [Babylon form of VariableDeclaration/VariableDeclarator](https://github.com/babel/babylon/blob/master/ast/spec.md#variabledeclaration). If this is not related to the Babylon AST, then perhaps I can make another change to the docs to make it clear where this difference from [ESTree](https://github.com/estree/estree/blob/master/es2015.md#variabledeclaration) comes from?

JamesHenry:
@ChristopherBiscardi Perhaps you could provide a concrete example?

AFAIK there are no deviations from ESTree in Babylon with regards to VariableDeclarations. The docs seems to confirm that: https://github.com/babel/babylon/tree/master#output

I also compared a simple declaration across acorn, espree and babylon and saw no difference in the relevant part of the ASTs.

What differences are you seeing?

ChristopherBiscardi:
ah, I suppose my confusion was based in the fact that there is no "current" ESTree markdown doc, only the `ES5.md` (which is the "base" spec), and three extensions (ES2015-2017), which means they are all accumulative and you have to read them all to understand any given structure.

closing this.

JamesHenry:
Cool, let me know if I can help out with any `prettier` PRs you are working on 👍

Pajn:
I'm of course also available if you have any questions about the parts where I have been

PR #207, opened 2017-03-20
Fix: Optional methods are not marked as optional (fixes #206)

### LINKED ISSUE #206: Optional methods are not marked as optional
These two produce the same AST
```
class A {
  optional?()
  required()
}

### PR DESCRIPTION
**Pull request by Pajn**

### CODE REVIEW

**Comment by soda0289** [comment]
I don't think you can have optional parameters defined in a class. This would only be valid when the class is declared
```ts
declare class A {
  optional?();
  required();
}
```

or maybe even an abstract class
```ts
abstract class A {
  public abstract optional?();
  
  public abstract required();
}
```

The typescript parser should throw an error not sure why its not.

soda0289:
I could be wrong thou it seems to work on typescript playground just seems odd to me.

Pajn:
If I understand the tests correctly optional properties are not allowed but optional methods are
<https://github.com/Microsoft/TypeScript/blob/ac67b94a6ea209760bfc31d9328337a16c41a0c1/tests/cases/conformance/types/namedTypes/classWithOptionalParameter.ts>
<https://github.com/Microsoft/TypeScript/blob/ac67b94a6ea209760bfc31d9328337a16c41a0c1/tests/cases/conformance/types/namedTypes/optionalMethods.ts>

However, non of the show errors in my editor so it seems I have stumbled upon a strange corner case :stuck_out_tongue:

soda0289:
I never knew of that typescript feature. I think we may want to add tests for the other two cases, declare and abstract, and maybe not have a class with a required function without a body since that should be an error in typescript.

soda0289:
I see the test case does not have a required function without a body. Never mind this should be good.

Pajn:
ah, sorry for that. I focused on the optional part and didn't realize the example in the issue was invalid. But as you noticed, that isn't the same as the test :)

I will add one ambient/declare and one abstract class as test cases as well.

PR #214, opened 2017-04-05
Chore: Wrap any parameter with modifers, not just in constructors

Pajn:
As pretter reprints the code from the AST, anything not in it will be gone.
So while
```
interface I {
    new (public x);
}
```
is invalid TypeScript, it's no good if pretter prints that as
```
interface I {
    new (x)
}
```
If you have your editor set up to format on save, the modifier will just
be gone and you'll never see the error from TypeScript and will therfore
consider it a bug.
Note that the code does not explicitly support this, I just moved the code
that supports this in constructor parameters so that it applies to any parameter.

As the code parses correctly with just missing information in the AST there
are no good way to know that the printed file will be different from the original.
So while there is no intention to support invalid TS it's easier to do that in this
case than to error.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1217,20 +1217,6 @@ module.exports = function(ast, extra) {
                                 return convertChild(d.expression);
                             }) : [];
 
-                            if (param.modifiers) {
-                                return {
-                                    type: "TSParameterProperty",
-                                    range: [param.getStart(), param.end],
-                                    loc: getLoc(param, ast),
-                                    accessibility: getTSNodeAccessibility(param),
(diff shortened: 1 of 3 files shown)


JamesHenry:
I think this is a grey area but is probably ok.

The reason I think that is that this is is not actually "structurally invalid" TypeScript. We can parse that source with no TS `parseDiagnostics` issues, and it produces a full TS AST.

Clearly, other parts of TS will then tell you that this is semantically invalid, but I think that is a level of abstraction above what we are focused on here.

JamesHenry:
P.S. @Pajn I changed this to a `Fix`, and fixed the typo in your commit ahead of merging.

PR #218, opened 2017-04-12
Breaking: Add .body to TSModuleBlock nodes (fixes #217)

### LINKED ISSUE #217: ImportDeclaration.parent is not expected to be a TSModuleDeclaration
the rule `import/newline-after-import` accesses a `ImportDeclaration`’s parent, expecting it to [have a `body`](https://github.com/benmosher/eslint-plugin-import/blob/6288cf93e73c3530845609d23fb829c379e357d8/src/rules/newline-after-import.js#L98)

that’s not true if you encounter a `TSModuleDeclaration` → `TSModuleBlock` → `ImportDeclaration`:

```ts
declare module 'postcss-modules' {
	import {Plugin} from 'postcss'
	...
}
```

a fix for this specific problem would be a visitor like `TSModuleDeclaration(m) { m.body = m.statements }`

flying-sheep:
I’ll probably have to fix some tests and don’t know if just choosing the name “ModuleBlock” is OK.

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1491,6 +1491,13 @@ module.exports = function(ast, extra) {
                 break;
 
             // Modules
+            case SyntaxKind.ModuleBlock:
+                assign(result, {
+                    type: "TSModuleBlock",
+                    body: node.statements.map(convertChild)
+                });
+                break;
+
             case SyntaxKind.ImportDeclaration:
                 assign(result, {
                     type: "ImportDeclaration",
--- a/tests/fixtures/typescript/namespaces-and-modules/ambient-module-declaration-with-import.result.js
+++ b/tests/fixtures/typescript/namespaces-and-modules/ambient-module-declaration-with-import.result.js
@@ -1,0 +1,349 @@
+module.exports = {
+    "body": [
+        {
+            "body": {
+                "body": [
+                    {
+                        "loc": {
+                            "end": {
+                                "column": 22,
+                                "line": 2
+                            },
+                            "start": {
+                                "column": 2,
+                                "line": 2
+                            }
+                        },
+                        "range": [
+                            34,
+                            54
+                        ],
+                        "source": {
+                            "loc": {
+                                "end": {
+                                    "column": 21,
+                                    "line": 2
+                                },
+                                "start": {
+                                    "column": 17,
+                                    "line": 2
+                                }
+                            },
+                            "range": [
+                                49,
+                                53
+                            ],
+                            "raw": "'fs'",
+                            "type": "Literal",
+                            "value": "fs"
+                        },
+                        "specifiers": [
+                            {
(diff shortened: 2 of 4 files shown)


soda0289:
Since the node is specific to typescript we should prefix it with TS. `TSModuleBlock` This is also a breaking change I think the commit message has to prefixed with `Breaking: `

soda0289:
@Pajn Will this change mess up the prettier work that has been completed?

flying-sheep:
OK, all done

Pajn:
>@Pajn Will this change mess up the prettier work that has been completed?

No, modules are not started yet so it's fine. Also more similarity with ESTree is nicer there as well so :+1:

soda0289:
@flying-sheep Looks like a namespace can also have import statements. We should update those as well.
```ts
declare namespace 'Baz' {
  import {Bar} from 'bar';
}
```

flying-sheep:
namespaces already convert to `TSModuleDeclaration` → `TSModuleBlock`.

please check the modified test file (`declare-namespace-with-exported-function.result.js`), which is the result of converting a namespace.

soda0289:
@flying-sheep Sorry missed that.

Is there a test for declared modules? I don't see any tests that failed. We should have one for declared modules to ensure we don't break anything in the future.

flying-sheep:
OK, done

JamesHenry:
Seems like we have a couple of breaking changes we can look to merge now, so we should get them in together

JamesHenry:
And another https://github.com/eslint/typescript-eslint-parser/pull/233 😄

azz:
I'm pretty sure `SyntaxKind.ModuleBlock` is legacy in name. The `namespace` keyword was originally `module`. They're now aliases, but `namespace` is recommended.

Since we have more freedom here, should we call the node `TSNamespaceBlock` instead?

JamesHenry:
@azz `module` and `namespace` are not always aliases

Being able to declare an ambient module name for the tsc is a useful feature for providing type information for 3rd party code. Namespace, however, cannot be used for this functionality.

E.g. https://www.typescriptlang.org/play/#src=declare%20module%20'lodash'%20%7B%20%7D%0D%0A%0D%0Adeclare%20namespace%20'lodash'%20%7B%7D

And AST confirmation (see the parseDiagnostics - those are parse errors):
http://astexplorer.net/#/gist/d63a43d869829f9a267517f036b4a22a/720b043dfff188b0aa1cadc1f55662a434efdb20

Reference for shorthand ambient module declarations: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#shorthand-ambient-module-declarations

azz:
@JamesHenry I was aware of the difference, but I was not aware that `declare module "x" {}` used the same AST nodes (`ModuleDeclaration` & `ModuleBlock`). Thanks.

PR #219, opened 2017-04-14
Breaking: Allow comment scanner to rescan tokens (fixes #216)

### LINKED ISSUE #216: Typescript finds comment inside of template string
2.2.1

2.1.0

```JS
const str = `${__dirname}/test/*.js`;

module.exports = str;
```

**What happened?**
```
Cannot read property 'loc' of null
TypeError: Cannot read property 'loc' of null
    at getNodeIndent (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/rules/indent.js:328:71)
    at checkNodeIndent (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/rules/indent.js:363:34)
    at nodes.forEach.node (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/rules/indent.js:411:35)
    at Array.forEach (native)
    at checkNodesIndent (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/rules/indent.js:411:19)
    at EventEmitter.Program (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/rules/indent.js:975:21)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:191:7)
    at NodeEventGenerator.applySelector (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/util/node-event-generator.js:265:26)
    at NodeEventGenerator.applySelectors (/Users/bradzacher/.config/yarn/global/node_modules/eslint/lib/util/node-event-generator.js:294:22)
```

note that these files seem to work fine, which leads me to believe it's a problem with the asterisk in the template string.
```JS
const str = `${__dirname}`;

module.exports = str;
```
```JS
const str = `${__dirname}/asdf.js`;

module.exports = str;
```

soda0289:
Template strings and regex patterns need to be rescaned to not scan tokens as comments. I aslo re-factored the comment code and removed the attachComment option since it is no longer needed.

The code pattern was taken from here:
https://github.com/Microsoft/TypeScript/blob/2b10611fbfb63c4fe24d865dc275fec016af003e/src/services/formatting/formattingScanner.ts

I'm not sure if there is a better way to find the container token node. We could use a list of scanned tokens then search through it for a possible container but I think we need to know the node and not the token to find a container. The method I use currently walks the ast searching for a container node.


soda0289:
@JamesHenry Can you give this a review. Maybe there is a better way to find a container token or we could use the typescript format scanner?

soda0289:
@JamesHenry its rebased now. I but the comment code in a separate file

azz:
Does the following parse correctly:

```
/* // test */
```

soda0289:
I added a test case for it. This is the result https://github.com/eslint/typescript-eslint-parser/blob/a5e2293d5d44c43b26b18e93eac27013ae0fe766/tests/fixtures/comments/line-comment-with-block-syntax.result.js

I changed the logic so like comments only remove //. Block comments only remove /* and */

lib/convert-comments.js
```diff
@@ -0,0 +1,147 @@
+/**
+ * @fileoverview Convert comment using TypeScript token scanner
+ * @author Nicholas C. Zakas
```
JamesHenry: Technically I wrote all the comment code 😄

lib/ast-converter.js
```diff
@@ -69,8 +70,8 @@ module.exports = (ast, extra) => {
     /**
      * Add the comment nodes to the AST (that were parsed separately in parser.js)
      */
-    if (extra.comment || extra.attachComment) {
-        estree.comments = extra.comments || [];
+    if (extra.comment) {
```
JamesHenry: I know this was redundant until now, but this is breaking change - prettier currently uses the `attachComment` version and will need to be updated.

I think we should still make the change, but we will need to update it to `Breaking`

JamesHenry:
This looks great @soda0289, thanks for all digging around `reScan` etc

soda0289:
@JamesHenry Thanks for the review.

Made the recommended changes and fixed a comment in parser.js.

PR #233, opened 2017-04-29
Breaking: Support TypeScript 2.3 (fixes #232)

### LINKED ISSUE #232: Support TypeScript 2.3
TypeScript 2.3 was recently released as the latest stable version and we should look to support it as soon as possible.

On initial inspection of a broken build, it looks like JSX will be the main area of focus for changes.

azz:
Seems to be a minor fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A parser that converts TypeScript into an [ESTree](https://github.com/estree/est
 
 We will always endeavor to support the latest stable version of TypeScript.
 
-The version of TypeScript currently supported by this parser is `~2.2.1`. This is reflected in the `devDependency` requirement within the package.json file, and it is what the tests will be run against. We have an open `peerDependency` requirement in order to allow for experimentation on newer/beta versions of TypeScript.
+The version of TypeScript currently supported by this parser is `~2.3.2`. This is reflected in the `devDependency` requirement within the package.json file, and it is what the tests will be run against. We have an open `peerDependency` requirement in order to allow for experimentation on newer/beta versions of TypeScript.
 
 If you use a non-supported version of TypeScript, the parser will log a warning to the console.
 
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks for the quick digging, @azz!

The fact that this didn't break any existing tests is a clear sign that we should add some as part of this change. Please can you add a relevant JSX test?

Many thanks!

JamesHenry:
Also, we would class this a `Breaking` change, and so the commit message needs to reflect that for our release process to work accordingly.

Reference: eslint.org/docs/developer-guide/contributing/pull-requests

azz:
I was just about to amend the commit to be backwards compatible. Is that something you want me to do?

JamesHenry:
Us changing the officially supported TypeScript version should basically always be a breaking change I think, in order to offer clear indication to our users that they will likely need to make changes to their setups

JamesHenry:
We have two other pending breaking changes to make, so this is great timing!

azz:
> The fact that this didn't break any existing tests is a clear sign that we should add some as part of this change. Please can you add a relevant JSX test?

Not sure what you mean by this. The tests did fail until this change, and the change itself it backwards compatible.

JamesHenry:
RE: The tests - Yes, sorry, I misread the type of change we were making at first.

In terms of backwards compatibility, you added that change and comment after I first looked at it. Please can you remove them again?

I really appreciate we you are trying to do with that, but if we added comments and "version bridges" for every TypeScript version upgrade, this project would become very difficult to manage. This upgrade is trivial, but previous ones were not at all. TypeScript does not follow a strict semver policy, and as mentioned before, the AST structure is naturally not respected at all when it comes to version numbers.

By classifying it as a breaking change, we will automatically release a new major version of the parser and the intention to our users is as clear as it can be.

azz:
> TypeScript does not follow a strict semver policy.

I learnt that the hard way when 2.1 came out...

> In terms of backwards compatibility, you added that change and comment after I first looked at it. Please can you remove them again?

Done.

JamesHenry:
We'll get this in with the other breaking changes and release a new major version 🎉

PR #241, opened 2017-04-29
Breaking: Change how interface node gets converted (fixes #201)

### LINKED ISSUE #201: Should Interfaces be converted?
I have started implementing support for interfaces in prettier but as the TS AST is a bit different than the ESTree it does not integrate as nice as other features have so far.

Flow does use `ObjectTypeAnnotation` for the body of the interface, similarly to how classes have a `ClassBody` node.

As the interface bodies can be quite different from my understanding (TS support a lot of (strange?) syntax in them). It might be best to keep TS prefixes on the type name even if we do this.

Just opening this for discussion, the current format is completely workable but it might at least be worth considering.

soda0289:
This commit create an TSInterfaceDeclration and TSInterfaceBody node when converting typescript interface

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -677,6 +677,25 @@ module.exports = function(ast, extra) {
         }
 
         /**
+         * Converts a child into a interface heritage node.
+         * @param {TSNode} child The TypeScript AST node to convert.
+         * @returns {ESTreeNode} The type annotation node.
+         */
+        function convertInterfaceHeritageClause(child) {
+            var id = convertChild(child.expression);
+            var classImplementsNode = {
+                type: "TSInterfaceHeritage",
+                loc: id.loc,
+                range: id.range,
+                id: id
+            };
+            if (child.typeArguments && child.typeArguments.length) {
+                classImplementsNode.typeParameters = convertTypeArgumentsToTypeParameters(child.typeArguments);
+            }
+            return classImplementsNode;
+        }
+
+        /**
          * For nodes that are copied directly from the TypeScript AST into
          * ESTree mostly as-is. The only difference is the addition of a type
          * property instead of a kind property. Recursively copies all children.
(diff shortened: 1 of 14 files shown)


azz:
Do we need `superTypeParameters` for interfaces too?

```ts
interface B<C, D> { }
interface A extends B<string, number> { }
```

soda0289:
I should change my tests to use extends. I thought interfaces implement other interfaces, but that's a mistake.

Maybe change the nose type to TSInterafceExtends?

tests/fixtures/typescript/basics/interface-implements.src.ts
```diff
@@ -0,0 +1,3 @@
+interface Foo implements Bar {
```
JamesHenry: It's not a parse error, but this is invalid TypeScript

> Interface declaration cannot have 'implements' clause.

tests/fixtures/typescript/basics/interface-with-implements-type-parameters.src.ts
```diff
@@ -0,0 +1,3 @@
+interface Foo<T> implements Bar<J> {
```
JamesHenry: See comment about this being invalid TypeScript

JamesHenry:
When this is updated to use extends, we could go for `superInterface` to align with `superClass`? There is no precedent in flow AFAIK...

JamesHenry:
On the whole, I definitely like the change, we may even want to look at refactoring the body in a future PR. Flow uses `ObjectTypeAnnotation` for that and a few other things. Not sure that existed when we started this

azz:
> When this is updated to use extends, we could go for superInterface to align with superClass? There is no precedent in flow AFAIK...

Can't. Unlike a class, an interface can extend many other interfaces or classes.

https://astexplorer.net/#/gist/5ef8ee37c06a2390c37aae0f9a61ed04/bf68403565ba586b351e1a46c73d1a4232c6cdca

JamesHenry:
How about `heritage: TSNode[]`?

Then we could even capture the semantically invalid `implement`s examples in the AST, as TypeScript itself would...

soda0289:
I'd be fine with using heritage. I'll also change the heritage node to TSInterfaceExtends but it could be TSInterfaceHeritage.

soda0289:
@JamesHenry I changed implements property to heritage.Also changed the node type to TSInterfaceHeritage

JamesHenry:
I'm pretty happy with this to go in now, thanks @soda0289! 😄 

Any other thoughts @azz?

azz:
LGTM :+1:

PR #243, opened 2017-04-30
Fix: Support superTypeParameters (fixes #242)

### LINKED ISSUE #242: superTypeParameters is missing in the AST

2.3.2

`master`.

```ts
class X extends Y<A, B> {}
```

Parse successfully, assigning to `superTypeParameters` in the AST.

**What happened?**

`superTypeParameters` is missing in AST.

azz:
Adds the `superTypeParameters` field to the `ClassExpression`/`ClassDeclaration` AST nodes.

prettier/prettier#1422

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -1550,6 +1550,9 @@ module.exports = function(ast, extra) {
 
                 if (hasExtends) {
                     superClass = heritageClauses.shift();
+                    if (superClass.types[0] && superClass.types[0].typeArguments) {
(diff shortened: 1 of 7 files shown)


soda0289:
Do you think we should have a new node for class extensions/superClass? That way we could keep the type parameter together with the identifier. Kind of like how we have a ClassImplements node

soda0289:
On second thought a new node for super class would mean we differ the estree  spec and might cause eslint rules to break.

azz:
Just pushed a new test with mixin classes. Does the AST look ok?

PR #246, opened 2017-05-02
Update semver package

SimenB:


soda0289:
Can you prefix the commit message with `Chore: `.

SimenB:
@soda0289 Went for `Upgrade`.

image

Want me to change it?

soda0289:
That's fine.
We should probably remove the need for every PR to depend on an issue. You can just ignore the bot.

PR #247, opened 2017-05-02
Chore: Update README with list of known issues

soda0289:
This commit updates the README with a list of known issues with ESLint rules and plugins.

@JamesHenry I think by providing a list of known issues with this parser we can help new users get started and cut back on duplicate issues. I don't mind when people open issues on failing rules or plugins as it helps determine the work needed to make the parser fully support eslint. Do you think this is a worth while change or might be better served by documentation?

@corbinu @mattdell  @flying-sheep Did I miss any known issues with this parser and eslint?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # TypeScript ESLint Parser (Experimental)
 
-A parser that converts TypeScript into an [ESTree](https://github.com/estree/estree)-compatible form so it can be used in ESLint. The goal is to allow TypeScript files to be parsed by ESLint (though not necessarily pass all ESLint rules).
+A parser that converts TypeScript into an [ESTree](https://github.com/estree/estree)-compatible form so it can be used in ESLint.
 
-**Important:** This parser is still in the very early stages and is considered experimental. There are likely a lot of bugs. You should not rely on this in a production environment yet.
+**Important:** This parser is not fully compatbile with all ESLint rules and plugins. Some rules will improperly mark source code as failing or not find problems where it should.
 
 ## Supported TypeScript Version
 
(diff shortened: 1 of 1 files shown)


Aladdin-ADD:
can we add known issues to https://github.com/eslint/typescript-eslint-parser/blob/master/.github/ISSUE_TEMPLATE.md?

soda0289:
@Aladdin-ADD We probably should. Not sure if a link to the README would be sufficient. Or if we should add more documentation with common work arounds too.

README.md
```diff
@@ -14,6 +14,23 @@ If you use a non-supported version of TypeScript, the parser will log a warning
...
+ - no-useless-constructor [#77](https://github.com/eslint/typescript-eslint-parser/issues/77)
+ - space-infix-ops [#224](https://github.com/eslint/typescript-eslint-parser/issues/224)
+ 
+The follow ESLint plugins have issues when used with this parser:
+ - eslint-plugin-react [#213](https://github.com/eslint/typescript-eslint-parser/issues/213)
+  
+The following TypeScript syntax will cause rules to fail or ESLint v3 to crash:
+ - Empty body functions
+   - Abstract methods
+   - Multiple function declarations
```
JamesHenry: Is this referring to function overloads?

soda0289: Yes function overloading. I can change the name to be more clear.

 I'm going to write up an issue for this problem. I think we might be able to come up with a better solution than the one I have proposed.

soda0289: https://github.com/eslint/typescript-eslint-parser/issues/253 Summarizes the current situation and the three PR's I created to fix this issue.

PR #251, opened 2017-05-04
Fix: add `instanceof` to ast-converter

despairblue:
The `BinaryExpression` created by `instanceof` does not set the operator of the node, because the keyword is missing in `TOKEN_TO_TEXT`.

This was discovered by https://github.com/prettier/prettier/issues/1480

fixes: #252

### DIFF
--- a/lib/ast-converter.js
+++ b/lib/ast-converter.js
@@ -59,6 +59,7 @@ TOKEN_TO_TEXT[SyntaxKind.GreaterThanEqualsToken] = ">=";
 TOKEN_TO_TEXT[SyntaxKind.EqualsEqualsToken] = "==";
 TOKEN_TO_TEXT[SyntaxKind.ExclamationEqualsToken] = "!=";
 TOKEN_TO_TEXT[SyntaxKind.EqualsEqualsEqualsToken] = "===";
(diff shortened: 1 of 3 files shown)


soda0289:
Thanks for the PR.

Can you add the prefix `Fix: ` or `New: ` to your commit.

I think @JamesHenry was working on fixing this is as well. I'll request a review from him just to make sure there is no issue with this change.

despairblue:
@soda0289 Thanks, all done.

soda0289:
@despairblue 
Could you move the test into the folder `tests/fixtures/basics` as it is not a typescript specific feature.

despairblue:
@soda0289 thanks, I think @eslintbot is happy now as well :)

platinumazure:
@despairblue If you remove the colon between "fixes" and the issue number, eslintbot will be happy. But we can also fix the commit message on merge, so it's up to you if you want to fix it on your end or not. Thanks!

despairblue:
@platinumazure 💡 thank you, done 😄

JamesHenry:
Yeah, that's super weird, I fixed this on a branch but apparently never submitted it as a PR 😄 I have no memory of either making the change nor why I didn't submit it!

PR #258, opened 2017-05-06
Fix: Do not mislabel JSXMemberExpression as JSXIdentifier (fixes #257)

### LINKED ISSUE #257: JSXMemberExpression incorrectly marked as a JSXIdentifier

2.3.2

```ts
var foo: Nested.Namespace.Type;
```

`JSXIdentifier` nodes should not have `property` and `object` keys. There is one in this AST that should be a `JSXMemberExpression`.

**What happened?**

```
{
  "type": "VariableDeclaration",
  "range": [
    0,
    30
  ],
  "loc": {
    "start": {
      "line": 1,
      "column": 0
    },
    "end": {
      "line": 1,
      "column": 30
    }
  },
  "declarations": [
    {
      "type": "VariableDeclarator",
      "range": [
        4,
        30
      ],
      "loc": {
        "start": {
          "line": 1,
          "column": 4
        },
        "end": {
          "line": 1,
          "column": 30
        }
      },
      "id": {
        "type": "Identifier",
        "range": [
          4,
          7
        ],
        "loc": {
          "start": {
            "line": 1,
            "column": 4
          },
          "end": {
            "line": 1,
            "column": 7
          }
        },
        "name": "foo",
        "typeAnnotation": {
          "type": "TypeAnnotation",
          "loc": {
            "start": {
              "line": 1,
              "column": 9
            },
            "end": {
              "line": 1,
              "column": 30
            }
          },
          "range": [
            9,
            30
          ],
          "typeAnnotation": {
            "type": "TSTypeReference",
            "range": [
              9,
              30
            ],
            "loc": {
              "start": {
                "line": 1,
                "column": 9
              },
              "end": {
                "line": 1,
                "column": 30
              }
            },
            "typeName": {
              "type": "JSXMemberExpression",
              "range": [
                9,
                30
              ],
              "loc": {
                "start": {
                  "line": 1,
                  "column": 9
                },
                "end": {
                  "line": 1,
                  "column": 30
                }
              },
              "object": {
                "type": "JSXIdentifier",
                "range": [
                  9,
                  25
                ],
                "loc": {
                  "start": {
                    "line": 1,
                    "column": 9
                  },
                  "end": {
                    "line": 1,
                    "column": 25
                  }
                },
                "object": {
                  "type": "JSXIdentifier",
                  "range": [
                    9,
                    15
                  ],
                  "loc": {
                    "start": {
                      "line": 1,
                      "column": 9
                    },
                    "end": {
                      "line": 1,
                      "column": 15
                    }
                  },
                  "name": "Nested"
                },
                "property": {
                  "type": "JSXIdentifier",
                  "range": [
                    16,
                    25
                  ],
                  "loc": {
                    "start": {
                      "line": 1,
                      "column": 16
                    },
                    "end": {
                      "line": 1,
                      "column": 25
                    }
                  },
                  "name": "Namespace"
                }
              },
              "property": {
                "type": "JSXIdentifier",
                "range": [
                  26,
                  30
                ],
                "loc": {
                  "start": {
                    "line": 1,
                    "column": 26
                  },
                  "end": {
                    "line": 1,
                    "column": 30
                  }
                },
                "name": "Type"
              }
            }
          }
        }
      },
      "init": null
    }
  ],
  "kind": "var"
}
```

azz:


JamesHenry:
@azz Sorry, I'm a bit confused about the test case you have included...

Namely, why is the source necessarily JSX-related at all?

You're `A.B.C` type could theoretically come from something like this:

```ts
namespace A {
    export namespace B {
        export type C = string
    }
}

var foo: A.B.C;
```

I think we might need to be more sophisticated with our work on `SyntaxKind.FirstNode`

soda0289:
I think typescript no longer uses `FirstNode` for JSX. I have disabled that code and none of the JSX test fail. @JamesHenry can you find a code example where JSX code uses `FirstNode`.

I think `FirstNode` is only used in type references when the Identifier has a period separator.

azz:
@JamesHenry I think I'm also confused. I was under the impression the JSXMemberExpression and JSXIdentifier were just re-used for TypeScript dotted property types (they actually print correctly in prettier).

Maybe it needs to be changed to `MemberExpression` and `Identifier` for `var x: A.B.C;`.

soda0289:
@azz I like the idea of using a MemberExpression. We could use that for both of these type annotations:
`const foo: a.b.c` and `const foo: a[b][c]`. 

Unless we want to come up with a new node type. We should define node types for all type annotations since currently I think we just append TS and don't have any handlers.

soda0289:
@azz If you have a change could you update the PR to match the babylon produced AST node discussed here:
https://github.com/JamesHenry/tsep-babylon-test/issues/12

azz:
@soda0289 Done. I've only changed `SyntaxKind.FirstNode`. Was there anything else?

PR #261, opened 2017-05-06
Chore: Refactor the codebase (fixes #220)

### LINKED ISSUE #220: Upgrade to ES6 and drop support for Node version < 4
Currently this parser still supports Node version 0.10 and uses ESLint v2 for linting. We should upgrade the codebase to ES6, using const and let, and stop using the assign library when the native `Object.assign()` can be used instead.

JamesHenry:
This is major refactoring of the structure of the core codebase.

**There are no changes to any existing test cases.**

Summary of changes:

- [x] Drop Node 0.12 support, now `>=4` to align with `eslint/eslint`
- [x] Remove now redundant `object-assign` dependency, and replace its usage with native `Object.assign`
- [x] Remove `features.js` and `token-translator.js` - they were never used
- [x] Use same lint configuration as `eslint/eslint`
- [x] Upgrade all dependencies to latest
- [x] Update `ast-node-types.js` to include all known converted `ESTreeNode` types
- [x] Add `errorOnUnknownASTType` parser option to throw if a generated `ESTreeNode` type is not found in `ast-node-types.js` (this is disabled by default, but enabled in tests)
- [x] Separate utility functions out into a new file, `node-utils.js`, to make the codebase much easier to reason about, especially w.r.t. which helpers mutate the current node, and which are pure functions
- [x] Move all `ts.*` helper functions into abstractions within `node-utils.js` so that it is clear what our dependencies on TypeScript actually are
- [x] Add brand new general utilities such as `findFirstMatchingChild` and `findFirstMatchingAncestor` which take in `predicate` functions, and are built upon by other utility functions. These also allow me to remove some `ts.*` dependencies from the codebase
- [x] Move core `convert()` function into its own file, `convert.js`, and increase what is explicitly passed into it rather than just closing over all outer scopes. Again, this greatly helps with clarity and focused responsibility
- [x] Add block scoping to more complex switch cases in `convert.js`, taking advantage of `let` and `const` for variable declarations. No more battling to come up with unique variable names within a 1000+ line function!


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-    - "0.12"
     - 4
     - 6
```
soda0289: Maybe add Node v7 as well?

lib/node-utils.js
```diff
@@ -0,0 +1,634 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Public
+//------------------------------------------------------------------------------
+
+/**
+ * Expose the enum of possible TSNode `kind`s.
+ */
+exports.SyntaxKind = SyntaxKind;
```
soda0289: Nit: Maybe we should move the exports to the bottom of the file and do it all in one object. That way it might be easier to know what a file exports. If this was typescript we could just export the function or variables but I don't care too much.

JamesHenry:
@soda0289 Yes fair point about the readability of exports, thanks! I have addressed that, and dialed up the linting strictness to fully match eslint/eslint (caught a lot of additional formatting points in test and Makefiles etc).

I updated the travis to run on 4, 5, 6 and 7, again to match eslint/eslint.

With regards to merging, I would love to get this in ASAP, and then continue working on more fixes. Are you happy to close out your old PRs and start branching off from this work?

soda0289:
Yea I can do that.

JamesHenry:
Thanks, sorry that this is awkward one - long overdue! I think I will get Pajn's breaking change one in first, then rebase this and merge it.

Then we could do an actual release I think, because all the breaking changes will be in.

soda0289:
Could we merge this into:
https://github.com/eslint/typescript-eslint-parser/pull/247 Prolly won't need a rebase.

I would like to get the comment scanner fixed before we release. Could you review that PR?

JamesHenry:
Both of those are in now, thanks!

soda0289:
Lets merge this then. I'll rebase the comment scanner code off of this change.

JamesHenry:
🎉

Issue #253, opened 2017-05-04
Empty body function cuase errors for several eslint rules


soda0289:
There was a long standing issue with this parser and escope that caused a crash when encountering an empty body function. This is because ESTree defines that all Function/Method Declarations and Function expressions should have a body. 

Escope has since been forked as eslint-scope and can now parse empty body function nodes. There are still serveral eslint rules that expect functions to have bodies.

Issues this caused are: #162 #92 #80 #78 

These issue have caused the parser to change the node type of the following function nodes:
 - Abstract method declaration - `TSAbstractMethods`
 - Function Declarations in namespace - `TSNamespaceFunctionDeclaration` (We do this even for non empty body functions which causes another problem #127)

There are other cases where we do not change the node type of empty body functions. These include:
 - Declared Classes
 - Declared Functions
 - Function overloading
 - Abstract Function expression (found within an abstract method declaration)

There are two options to fix this. We can change the type of these node or modify the failing eslint rules.

I have proposed the following PRs: #166 #165 #163 

All of these change the type of both the function declaration and function expression. Maybe we should just remove the function expression entierly. We can create the following new nodes:

```
TSAbstractMethodDefinition {
  id: Identifier;
  params: [ Parameters ];
  returnType: TypeAnnotation;
  typeParameters: [ TSTypeParameter ];
}
```

```
TSAmbientFunctionDefinition {
  id: Identifier;
  params: [ Parameters ];
  returnType: TypeAnnotation;
  typeParameters: [ TSTypeParameter ];
  declared: boolean;
}
```

```
TSOverloadFunctionDefinition {
  id: Identifier;
  params: [ Parameters ];
  returnType: TypeAnnotation;
  typeParameters: [ TSTypeParameter ];
}
```

Sorry about the long post but wanted to create a summary of the current situation and allow for easier discussion. If we keep keep the nodes the way they are we might want to change back the TSNamespaceFunction declaration since it is causing another issue.

@JamesHenry thoughts?

soda0289:
@JamesHenry Any change we could a fix in for this in the next release.
The follow eslint rules will fail currently:
 - indent v4
 - array-callback-return
 - lines-around-directive
 - unicorn/custom-error-definition

ismail-syed:
👋 @JamesHenry @soda0289, is this issue being worked on? I'm having a related issue that I've described [here](https://github.com/eslint/typescript-eslint-parser/issues/92#issuecomment-328929560).

[...]

PR #264, opened 2017-05-07
Breaking: Convert Signature types to be more ESTree like (fixes #262)

### LINKED ISSUE #262: TSPropertySignature has no computed flag

2.3.2

```
var x: { [y]: any };
```

`TSPropertySignature` to have a `computed: true` property.

**What happened?**

No `computed` property.

In prettier, we currently print this as `var x: { y: any }`, which is incorrect.

prettier/prettier#1480

soda0289:
I noticed that optional is used twice. Once in identifier and once in the TSPropertySignature. Maybe we should clean that up.

Here is what this produces:
https://github.com/eslint/typescript-eslint-parser/issues/262#issuecomment-299722772

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -618,7 +618,7 @@ module.exports = function convert(config) {
                 type: AST_NODE_TYPES.Property,
                 key: convertChild(node.name),
                 value: convertChild(node.initializer),
-                computed: (node.name.kind === SyntaxKind.ComputedPropertyName),
+                computed: nodeUtils.isComputedProperty(node.name),
                 method: false,
                 shorthand: false,
                 kind: "init"
@@ -660,7 +660,7 @@ module.exports = function convert(config) {
                 type: (isAbstract) ? AST_NODE_TYPES.TSAbstractClassProperty : AST_NODE_TYPES.ClassProperty,
                 key: convertChild(node.name),
                 value: convertChild(node.initializer),
-                computed: (node.name.kind === SyntaxKind.ComputedPropertyName),
+                computed: nodeUtils.isComputedProperty(node.name),
                 static: nodeUtils.hasStaticModifierFlag(node),
                 accessibility: nodeUtils.getTSNodeAccessibility(node),
                 decorators: convertDecorators(node.decorators),
@@ -713,7 +713,7 @@ module.exports = function convert(config) {
                     type: AST_NODE_TYPES.Property,
                     key: convertChild(node.name),
                     value: method,
-                    computed: (node.name.kind === SyntaxKind.ComputedPropertyName),
+                    computed: nodeUtils.isComputedProperty(node.name),
                     method: nodeIsMethod,
                     shorthand: false,
                     kind: "init"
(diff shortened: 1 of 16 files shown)


soda0289:
We can move optional and computed to the Identifier node. I think that might make more sense

tests/fixtures/typescript/basics/interface-with-all-property-types.src.ts
```diff
@@ -0,0 +1,15 @@
+interface Foo {
```
JamesHenry: A number of these aren't semantically valid TypeScript, but I guess we are further evaluating errorRecovery here, so it should be ok

lib/convert.js
```diff
@@ -1688,6 +1688,59 @@ module.exports = function convert(config) {
 
         }
 
+        case SyntaxKind.MethodSignature: {
+            Object.assign(result, {
+                type: AST_NODE_TYPES.TSMethodSignature,
+                optional: (node.questionToken) ? (node.questionToken.kind === SyntaxKind.QuestionToken) : false,
```
JamesHenry: Worth a util function?

lib/convert.js
```diff
@@ -1688,6 +1688,59 @@ module.exports = function convert(config) {
...
+                computed: node.name.kind === SyntaxKind.ComputedPropertyName,
```
JamesHenry: Worth a util function?

soda0289:
@JamesHenry Thanks for the review. Will make the changes.

The only other breaking changes we could do are:
 - Change how we label `declare` variable, functions, and classes. I think prettier no longer depends on this since they use the source code to check if the keyword is present but we should still have a better solution for the long term.
 - If we implement this a fix for #253 it would be a breaking change but not sure if we could just patch all the eslint rules, even thou some are in plugins.
 - There are these two requests but they break current spec #266 #267

soda0289:
I added two functions to node-utils:
 - `isComputedProperty()`
 - `isOptional()`

tests/fixtures/typescript/basics/interface-with-all-property-types.src.ts
soda0289: I can split them into two test. Not sure which ones are invalid exactly but I think the computed property don't make sense in interfaces.

JamesHenry:
Again, just want to check how/if the AST alignment work affects this?

soda0289:
I need to change this so that only `optional` and `typeAnnoations` appear on Identifier node. Everything else should be on parent node.

JamesHenry:
I think we should look to finalise this after https://github.com/eslint/typescript-eslint-parser/pull/292 gets resolved right?

soda0289:
@JamesHenry Should be good.

JamesHenry:
This LGTM, @vjeux there are a lot of changes here, would you mind just confirming that this is all what you are expecting?

vjeux:
Yup, I actually made prettier use this branch :) Shipit!

JamesHenry:
Thanks, chaps! Awesome work

PR #270, opened 2017-05-13
fix: wrap interface in ExportNamedDeclaration if necessary (fixes #269

### LINKED ISSUE #269: abstract/export keyword missing from AST for interfaces
2.3.2

f5fcc879b493138eb97b9bdfb39afc4d210b1ddf

```ts
abstract interface I {}
```

I expected the AST to contain the information that the interface is abstract.

**What happened?**
It's not in the AST

```js
module.exports = {
    type: "Program",
    range: [0, 23],
    loc: {
        start: {
            line: 1,
            column: 0
        },
        end: {
            line: 1,
            column: 23
        }
    },
    body: [
        {
            type: "TSInterfaceDeclaration",
            range: [0, 23],
            loc: {
                start: {
                    line: 1,
                    column: 0
                },
                end: {
                    line: 1,
                    column: 23
                }
            },
            body: {
                type: "TSInterfaceBody",
                body: [],
                range: [21, 23],
                loc: {
                    start: {
                        line: 1,
                        column: 21
                    },
                    end: {
                        line: 1,
                        column: 23
                    }
                }
            },
            id: {
                type: "Identifier",
                range: [19, 20],
                loc: {
                    start: {
                        line: 1,
                        column: 19
                    },
                    end: {
                        line: 1,
                        column: 20
                    }
                },
                name: "I"
            },
            heritage: []
        }
    ],
    sourceType: "script",
    tokens: [
        {
            type: "Identifier",
            value: "abstract",
            range: [0, 8],
            loc: {
                start: {
                    line: 1,
                    column: 0
                },
                end: {
                    line: 1,
                    column: 8
                }
            }
        },
        {
            type: "Keyword",
            value: "interface",
            range: [9, 18],
            loc: {
                start: {
                    line: 1,
                    column: 9
                },
                end: {
                    line: 1,
                    column: 18
                }
            }
        },
        {
            type: "Identifier",
            value: "I",
            range: [19, 20],
            loc: {
                start: {
                    line: 1,
                    column: 19
                },
                end: {
                    line: 1,
                    column: 20
                }
            }
        },
        {
            type: "Punctuator",
            value: "{",
            range: [21, 22],
            loc: {
                start: {
                    line: 1,
                    column: 21
                },
                end: {
                    line: 1,
                    column: 22
                }
            }
        },
        {
            type: "Punctuator",
            value: "}",
            range: [22, 23],
            loc: {
                start: {
                    line: 1,
                    column: 22
                },
                end: {
                    line: 1,
                    column: 23
                }
            }
        }
    ]
};
```

despairblue:


despairblue:
@soda0289 done.

The diffs for the other tests are a bit bigger because I used `tools/update-typescript-tests.js` to generate them. I also fixed `tools/update-typescript-tests.js` to use the same `getRaw` implementation the tests use and added a missing `shelljs.echo` call.

Abstract interfaces are not valid TypeScript btw. But I think they added the test to make sure the parser can handle that, so they can output a useful error message like `'abstract' modifier can only appear on a class, method, or property declaration.` instead of `Expected ClASS or STRING. Instead found ABSTRACT`.

soda0289:
@despairblue Thanks for making the changes.

If abstract interfaces are not valid typescript I'm not sure we should support them or the very least only set `abstract: true` when it is present and not set `abstract: false`. I'll let @JamesHenry just to make sure this change makes sense.

despairblue:
I'm happy with whatever choice you make.

That said, my understanding is that this is an AST to AST converter, so it should convert the AST as accurately as possible and should not make assumptions about it's semantics as a program.

But again, it's your choice, we just would need to blacklist some tests from the typescript test suite, no biggie :)

despairblue:
@soda0289 @JamesHenry is there anything I can do to help this get in? I could remove the `abstract` if that's the problem.

soda0289:
I'm fine with using abstract. I think we are also going to change classes to use a property `abstract` now as well instead of a new node type. I will merge this soon.

soda0289:
@despairblue There will be quite a few changes coming to the AST as we try to align with the AST from the new typescript babylon plugin. You can check the issues here:
https://github.com/JamesHenry/tsep-babylon-test/issues

PR #271, opened 2017-05-14
Fix: Use correct starting range and loc for JSXText tokens (fixes #227)

### LINKED ISSUE #227: Linting JSX const without brackets / parenthesis causes eslint to hang
2.2.1

2.1.0

```
const menuItems = <div className={theme['header__menu-items']}>
        <Icon handleClick={handleIconClick} type={IconTypes.Chat} />
        <Icon handleClick={handleIconClick} type={IconTypes.Menu} />
    </div>;
```

eslint to run successfully

**What happened?**
eslint hangs and becomes unresponsive. I've left it to run over lunch and it was still stuck.

Changing the syntax to be contained within brackets / parenthesis fixes the issue.

```
const menuItems = (
    <div className={theme['header__menu-items']}>
        <Icon handleClick={handleIconClick} type={IconTypes.Chat} />
        <Icon handleClick={handleIconClick} type={IconTypes.Menu} />
    </div>
);
```

 I've now been able to reproduce this several times.

soda0289:
We used to use the method `getStart()` to calculate the starting range for JSXText tokens. This would skip over whitespace and cause the value of these tokens to be empty. This would also result in the range and starting line number and column to be incorrect. This commit uses the method `getFullStart()`, on JSXText tokens, which does not skip over whitespace and allows for the correct range, value and loc to be set.

This also fixes issues with JSX indentation in the new indent rule and in the eslint-plugin-react.

### DIFF
--- a/lib/node-utils.js
+++ b/lib/node-utils.js
@@ -588,15 +588,16 @@ function getTokenType(token) {
  * @returns {ESTreeToken}       the converted ESTreeToken
  */
 function convertToken(token, ast) {
-    const start = token.getStart(),
-        value = ast.text.slice(start, token.end),
+    const start = (token.kind === SyntaxKind.JsxText) ? token.getFullStart() : token.getStart(),
+        end = token.getEnd(),
+        value = ast.text.slice(start, end),
         newToken = {
             type: getTokenType(token),
             value,
             start,
-            end: token.end,
-            range: [start, token.end],
-            loc: getLoc(token, ast)
+            end,
+            range: [start, end],
+            loc: getLocFor(start, end, ast)
         };
 
     if (newToken.type === "RegularExpression") {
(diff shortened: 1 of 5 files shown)


lib/node-utils.js
```diff
@@ -588,15 +588,16 @@ function getTokenType(token) {
  * @returns {ESTreeToken}       the converted ESTreeToken
  */
 function convertToken(token, ast) {
-    const start = token.getStart(),
+    const start = (token.kind === SyntaxKind.JsxText) ? token.getFullStart() : token.getStart(),
+        end = token.getEnd(),
         value = ast.text.slice(start, token.end),
```
JamesHenry: All the other references to `token.end` were updated to the variable `end`, but not this one. Just checking that is intended/not significant?

soda0289: I was reading through the code of typescript and noticed this method. I thought it looked a little cleaner since we use methods to access the starting point.

These are the methods to access token positions:
 - `getFullStart()` is the same as `.pos`
 - `getEnd()` is the same as `.end`
 - `getStart()` is the same as `.pos` but skipping over trivia and white space

If you want I can change the logic to use the properties to be more consistent with the rest of the code.

PR #272, opened 2017-05-15
New: Create option to enable JSXText node type (fixes #266)

### LINKED ISSUE #266: JSXText nodes are mislabeled as Literal nodes

2.3.2

```jsx
<a> </a>
```

`Program.body[0].expression.children[0]` should be a `JSXText` node.

**What happened?**

`Program.body[0].expression.children[0]` is a `Literal` node.

```js
{ type: 'Literal',
  range: [ 3, 4 ],
  loc: { start: { line: 1, column: 3 }, end: { line: 1, column: 4 } },
  value: ' ',
  raw: ' ' }
```

prettier/prettier#1558

soda0289:
This commit creates a new parser config option `useJSXTextNode` that enables `JSXText` node type instead of `Literal` for strings inside of `JSXElement` nodes. The default is false as currently many eslint rules depend on the `Literal` node type. 

I have moved the JSX tests into its own directory and created a new test runner that can run tests for JSX files with and without the  `useJSXTextNode`  flag enabled.


tests/lib/jsx.js
```diff
@@ -0,0 +1,110 @@
...
+//------------------------------------------------------------------------------
+
+const JSX_FIXTURES_DIR = "./tests/fixtures/jsx";
+
+const filesWithOutsandingTSIssues = [
+    "jsx/embedded-tags", // https://github.com/Microsoft/TypeScript/issues/7410
+    "jsx/namespaced-attribute-and-value-inserted", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/namespaced-name-and-attribute", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/invalid-namespace-value-with-dots", // https://github.com/Microsoft/TypeScript/issues/7411
+    "jsx/multiple-blank-spaces" // Fixed by #271
```
JamesHenry: This was fixed, why has it been re-added as one to skip?

lib/convert.js
```diff
@@ -1613,9 +1613,12 @@ module.exports = function convert(config) {
 
         }
 
-        case SyntaxKind.JsxText:
+        case SyntaxKind.JsxText: {
```
JamesHenry: Please can you document inline why we are doing this?

soda0289: Sure. It is only temporary until eslint v5, will make a note.

tests/lib/jsx.js
soda0289: Once I rebase this change I can remove it.

JamesHenry:
Ah, I just merged the other JSXText PR, so this will need rebasing. Once that's done this is good to go, and I can see it has already been verified in prettier.

soda0289:
@JamesHenry Rebased. Should be good to go in.

PR #277, opened 2017-05-20
Fix: Convert range and line number corretly in JSX literals

soda0289:
I ony fixed the JSXText tokens in my last commit. The Literal node needs to be fixed to use the correct range and line number. This fixes more issue with eslint-plugin-react rules


OliverJAsh:
Any ideas when will this be released?

PR #279, opened 2017-05-20
Fix: Set node type to ExperimentalRestProperty (fixes #276)

### LINKED ISSUE #276: False positive on `comma-dangle` with `RestProperty`
`2.3.2`

`3.0.0`

Code is closed-source, but the gist is this:
```
const SomeComponent = withStyles(styles)(({
  children,
  ...props
}) => {
```

No warning for missing comma at the `...props` line.

**What happened?**
`warning  Missing trailing comma  comma-dangle` for the `...props` line.

Eslint recently disabled the trailing comma after the "RestProperty" but the parser likely generates "Property" node (based on my hacked-in logging from within the `comma-dangle` rule).

This warning: https://github.com/eslint/eslint/blob/70367d1a48c8c71f9fbde05da4a7a626a59967fd/lib/rules/comma-dangle.js#L266

Here's the check for "RestProperty" which seems to fail because just "Property" is there instead:
https://github.com/eslint/eslint/blob/70367d1a48c8c71f9fbde05da4a7a626a59967fd/lib/rules/comma-dangle.js#L34-L40

soda0289:
We were not using the correct node type when an object spread was used in a function parameter. This commit checks for the dotDotDot property can creates an `ExperimentalRestPropert` if set. ~I don't think typescript supports using an initializer or property name so I do not convert these properties.~ Edit: TypeScript does support inital value and renaming of property. Will update the PR and tests


JamesHenry:
So, this is not yet ready for review?

soda0289:
@JamesHenry should be good to review. I copied the object rest spread tests from espree and made some changes to allow all the tests to pass.

PR #293, opened 2017-05-25
Breaking: Decorator ESTree compliance, always convert (fixes #250)

### LINKED ISSUE #250: Decorator output is inconsistent
In this example, the decorator is not wrapped in a `Decorator` node:

```js
class Person {
  @inject leftLeg: Leg;
}

==>

node.decorators = [ {type: 'Identifier'} ];
```

In this example, it is wrapped in a `TSDecorator` node, instead of `Decorator` node:

```js
@dec enum E {}

==>

node.decorators = [ {type: 'TSDecorator', expression: {type: 'Identifier'}} ];
```

It would be nice if decorators arrays was always wrapped in a `Decorator` node (without `TS` prefix). This way it would work out of the box with babylon decorators.

I have a workaround in prettier, but I'd rather it be fixed inside of this project: https://github.com/prettier/prettier/pull/1509

JamesHenry:
(convertDecorators had to be moved purely to comply with "no use before define" rule)


JamesHenry:
Ah just noticed the second part of that issue - wrapping in a Decorator node. This only deals with the consistency of when we convert, I will address _how_ we convert too.

JamesHenry:
I have updated the PR to both fix our ESTree compliance, and ensure conversion during `deeplyCopy()`.

Importantly, I also took this opportunity to "fix" `tools/update-typescript-tests.js` so that it is actually useful 😄 

https://github.com/eslint/typescript-eslint-parser/pull/293/files#diff-7074ce236b7ec2f768cbd30e59304265

It will now use the same assertion logic as the tests, and only perform a write to an existing result file if the assert fails.

tools/update-typescript-tests.js
```diff
@@ -26,50 +21,60 @@ var shelljs = require("shelljs"),
...
 
+/**
+ * Writes the given AST to the given filename
+ * @param {*} result the AST to write to a file
+ * @param {*} testResultFilename name of the file to write to
+ * @returns {void}
+ */
 function outputResult(result, testResultFilename) {
-    shelljs.echo("module.exports = " + JSON.stringify(result, null, "    ") + ";").to(testResultFilename);
+    shelljs.echo(`module.exports = ${JSON.stringify(result, null, "    ")};`).to(testResultFilename);
```
soda0289: I thought about using this tool but sometimes it will generate JSON objects that are inconsistent or change for no reason. By adding the deep equal assertion it does prevent updating every file but maybe we should make the stringifyed JSON output consistent. We can use a tool like this: https://github.com/substack/json-stable-stringify to sort the output keys. This could be done later just a thought.

JamesHenry:
Rebased this, will merge once it has built

PR #300, opened 2017-05-28
Chore: Replace mocha (istanbul, chai, leche) with Jest

JamesHenry:
This PR exactly matches our current usage of Mocha (+ istanbul, chai and leche) but using Jest.

There is no change in how assertions are being made, nor how coverage is reported (Jest can provide it out of the box).

I have continued to use the original Makefile for the primary `npm test` command so as to minimise the changes in this initial PR, but added an additional `npm run jest` command (for development purposes), because Jest gives really nice output for failures/diffs and you lose the colour from that when executing it via another tool (Makefile nodeCLI).

Jest's watch mode is also excellent for development, so you now easily run that using `npm run jest -- --watch`.

I plan to explore formatted snapshot testing with Jest in a follow up PR, because it will provide an effortless way for us to deterministically update test cases when we make breaking changes (which will we be doing a lot more of to align with the Babylon work)


soda0289:
Should we merge this in for the next release?

JamesHenry:
Shouldn't make any difference either way, it's a development-only concern. I'm happy to wait until the next batch

vjeux:
> I plan to explore formatted snapshot testing with Jest in a follow up PR

This has been invaluable for prettier. I honestly don't think I would have finished it without snapshot testing. I highly recommend :)

PR #311, opened 2017-06-05
Chore: Refactor tests to assert snapshots not JSON

JamesHenry:
Naturally, this is a huge diff, but brings with it all of the benefits previously discussed.

- I have so far only created the snapshots and not yet deleted the old `result.js` files, purely to make this easier to review and reduce noise. To be clear, in this PR the `result.js` files are no longer being used, and I plan to remove them as soon as you are happy with the rest of the changes.
- I was able to greatly simplify the execution of the assertions now that it relies on consistent tooling (see the `createSnapshotTestBlock()` helper function below).
- The unreliable `update-typescript-tests` script is no longer necessary, as jest provides an effortless way to update the snapshots out of the box.


JamesHenry:
Thanks, @soda0289! Updated to include the removals. As well as all the `.result.js` files, we actually had a lot of unused fixtures in there, so I have removed those too. I will merge when it successfully builds.

PR #313, opened 2017-06-06
Fix: Use TSNullKeyword for null type instead of Literal

JamesHenry:
I went for the simplest possible solution for now to resolve https://github.com/JamesHenry/tsep-babylon-test/issues/16.

Another awkward one in terms of Breaking vs Fix, but Fix feels ok here.


JamesHenry:
P.S. Developing this in jest watch mode with the snapshot diff was a joy 😄

JamesHenry:
I think this needs a more holistic solution, already found a case that this change won't catch...

JamesHenry:
Updated with a much better fix, now ready for review

PR #319, opened 2017-06-09
Breaking: Include type annotation range for Identifiers (fixes #314)

### LINKED ISSUE #314: Incorrect location for typeAnnotation
806251524424e3ad91e750da0f76b8de25ed0b42

```js
function x(o: { x: string }) {}
```

I expect that the `Identifier` location to include its typeAnnotation location: 11 - 27.

We do a binary search in order to find the node nearest a comment and because the location is not inclusive, we don't go inside of the identifier and don't attach it correctly :( https://github.com/prettier/prettier/issues/1946

**What happened?**

Right now, it gives me 11 - 12.

```js
      "params": [
        {
          "type": "Identifier",
          "range": [
            11, // <---------
            12 // <---------
          ],
          "loc": {
            "start": {
              "line": 1,
              "column": 11
            },
            "end": {
              "line": 1,
              "column": 12
            }
          },
          "name": "o",
          "typeAnnotation": {
            "type": "TypeAnnotation",
            "loc": {
              "start": {
                "line": 1,
                "column": 14
              },
              "end": {
                "line": 1,
                "column": 27
              }
            },
            "range": [
              14, // <---------
              27 // <---------
            ],
            "typeAnnotation": {
              "type": "TSTypeLiteral",
              "range": [
                14,
                27
              ],
              "loc": {
                "start": {
                  "line": 1,
                  "column": 14
                },
                "end": {
                  "line": 1,
                  "column": 27
                }
              },
              "members": [
                {
                  "type": "TSPropertySignature",
                  "range": [
                    16,
                    25
                  ],
                  "loc": {
                    "start": {
                      "line": 1,
                      "column": 16
                    },
                    "end": {
                      "line": 1,
                      "column": 25
                    }
                  },
                  "optional": false,
                  "computed": false,
                  "key": {
                    "type": "Identifier",
                    "range": [
                      16,
                      17
                    ],
                    "loc": {
                      "start": {
                        "line": 1,
                        "column": 16
                      },
                      "end": {
                        "line": 1,
                        "column": 17
                      }
                    },
                    "name": "x"
                  },
                  "typeAnnotation": {
                    "type": "TypeAnnotation",
                    "loc": {
                      "start": {
                        "line": 1,
                        "column": 19
                      },
                      "end": {
                        "line": 1,
                        "column": 25
                      }
                    },
                    "range": [
                      19,
                      25
                    ],
                    "typeAnnotation": {
                      "type": "TSStringKeyword",
                      "range": [
                        19,
                        25
                      ],
                      "loc": {
                        "start": {
                          "line": 1,
                          "column": 19
                        },
                        "end": {
                          "line": 1,
                          "column": 25
                        }
                      }
                    }
                  },
                  "initializer": null,
                  "accessibility": null,
                  "readonly": false,
                  "static": false,
                  "export": false
                }
              ]
            }
```

soda0289:
I changed range of Identifier nodes used in variable declaration and function parameters. Its a simple change but I might have missed something.

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -535,6 +535,7 @@ module.exports = function convert(config) {
 
             if (node.type) {
                 result.id.typeAnnotation = convertTypeAnnotation(node.type);
+                result.id.range[1] = node.type.getEnd();
             }
             break;
 
@@ -1145,9 +1146,8 @@ module.exports = function convert(config) {
             }
 
             if (node.type) {
-                Object.assign(parameter, {
-                    typeAnnotation: convertTypeAnnotation(node.type)
-                });
+                parameter.typeAnnotation = convertTypeAnnotation(node.type);
+                parameter.range[1] = node.type.getEnd();
             }
 
             if (node.questionToken) {
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -526,7 +526,7 @@ module.exports = function convert(config) {
 
         }
 
-        case SyntaxKind.VariableDeclaration:
+        case SyntaxKind.VariableDeclaration: {
             Object.assign(result, {
                 type: AST_NODE_TYPES.VariableDeclarator,
                 id: convertChild(node.name),
@@ -536,8 +536,17 @@ module.exports = function convert(config) {
             if (node.type) {
                 result.id.typeAnnotation = convertTypeAnnotation(node.type);
                 result.id.range[1] = node.type.getEnd();
+
+                const identifierEnd = node.name.getEnd();
+                const numCharsBetweenTypeAndIdentifier = node.type.getStart() - (node.type.getFullStart() - identifierEnd - ":".length) - identifierEnd;
+
+                result.id.typeAnnotation.range = [
+                    result.id.typeAnnotation.range[0] - numCharsBetweenTypeAndIdentifier,
+                    result.id.typeAnnotation.range[1]
+                ];
             }
             break;
+        }
 
         case SyntaxKind.VariableStatement:
             Object.assign(result, {
--- a/tests/fixtures/typescript/basics/var-with-type.src.ts
+++ b/tests/fixtures/typescript/basics/var-with-type.src.ts
@@ -1,1 +1,2 @@
 var name:string = "Nicholas";
+var foo: string = "Bar";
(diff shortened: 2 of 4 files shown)


JamesHenry:
I got chance to dig into this today, and we still have a fundamental difference in how we conclude ranges for typeAnnotations vs how Flow does it for the same source.

Take this example:

```ts
let x   :      string;
```

Flow has the following ranges:
screen shot 2017-06-18 at 18 22 43

Whereas in this PR we currently have:

screen shot 2017-06-18 at 18 30 01


I have a fix for this when it comes to `VariableDeclaration`s (building on this PR) so I'll just push it up for your review @soda0289

JamesHenry:
I rebased against master as it was a little bit behind

soda0289:
Thanks for looking into this. Just to confirm the change now uses the colon token as the initial start value of the type annotation node.

JamesHenry:
@soda0289 Yes, purely based on Flow's precedent of doing that

JamesHenry:
Based on a really quick check, it seems like this change will not require any updates to prettier which is a bonus.

JamesHenry:
@andy-ms Please could you take a look at this also? This would be a solution to https://github.com/JamesHenry/tsep-babylon-test/issues/19

See above for the precedent set by flow, I think it differs slightly from babylon's current output

soda0289:
I have not tested this with eslint yet, not sure if any rules will fail. Will try it out tonight

JamesHenry:
@soda0289 Did you get chance to try it out? How did/were you planning on evaluating it out of interest? Do you have a particular repo set up?

soda0289:
Sorry forgot about this. Was a little busy last week. I have several repos at work that I can run this change under and make sure there are no crashes or false positives.

soda0289:
@JamesHenry 
This does cause an issue with the eslint-plugin-typescript. The rule  typescript/type-annotation-spacing has false positives. Which makes sense since we are changing how the type annotation range is calculated.

JamesHenry:
I think that plugin currently has a number of issues right? I would think it should be pretty straightforward to fix

soda0289:
Yes the plugin does have several issues. Should be good to merge this too.

PR #347, opened 2017-08-08
Only warn about an unsupported TypeScript version once

### LINKED ISSUE #348: When using an unsupported TypeScript version, the console gets spammed with warning messages
2.4.1

4.0.0

A project with lots of files

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-eslint-parser.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: ~2.3.2

YOUR TYPESCRIPT VERSION: 2.4.1

Please only submit bug reports when using the officially supported version.

=============
```

**What happened?**

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-eslint-parser.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: ~2.3.2

YOUR TYPESCRIPT VERSION: 2.4.1

Please only submit bug reports when using the officially supported version.

=============
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-eslint-parser.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: ~2.3.2

YOUR TYPESCRIPT VERSION: 2.4.1

Please only submit bug reports when using the officially supported version.

=============
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-eslint-parser.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: ~2.3.2

YOUR TYPESCRIPT VERSION: 2.4.1

Please only submit bug reports when using the officially supported version.

=============
```
(You get the idea)

j-f1:
It gets annoying when you’re trying to lint a project with lots of files. Fixes #348.

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -100,6 +100,9 @@ function parse(code, options) {
          */
         if (typeof options.loggerFn === "function") {
             extra.log = options.loggerFn;
+        } else if (options.loggerFn === false) {
+            /** */ // eslint-disable-line valid-jsdoc
+            extra.log = () => {};
         }
 
     }
--- a/parser.js
+++ b/parser.js
@@ -27,7 +27,7 @@ if (!isRunningSupportedTypeScriptVersion) {
         "Please only submit bug reports when using the officially supported version.",
         border
     ];
-    console.log(versionWarning.join("\n\n"));
+    console.log(versionWarning.join("\n\n")); // eslint-disable-line no-console
 }
 
 let extra;
--- a/parser.js
+++ b/parser.js
@@ -16,21 +16,8 @@ const SUPPORTED_TYPESCRIPT_VERSIONS = require("./package.json").devDependencies.
 const ACTIVE_TYPESCRIPT_VERSION = ts.version;
 const isRunningSupportedTypeScriptVersion = semver.satisfies(ACTIVE_TYPESCRIPT_VERSION, SUPPORTED_TYPESCRIPT_VERSIONS);
 
-if (!isRunningSupportedTypeScriptVersion) {
-    const border = "=============";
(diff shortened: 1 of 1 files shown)


soda0289:
I think we can move the warning text outside the `parse()` that should eliminate the use of a global variable

j-f1:
@soda0289 Done!

soda0289:
@j-f1
 Thanks for making the change but on a second look I noticed we need to have access to the extra.log() and dont think this should be outside parse(). That was added since we did not want to write logs when other projects that use this parser. We still need to have the ability for users of the library to overwrite the logger.

I think we will have to use the global variable. We should move the warning text to its own function and contain the logic of checking if it has been called to there. Sorry for the confusion and thanks for the PR.

parser.js
```diff
@@ -99,11 +100,14 @@ function parse(code, options) {
          */
         if (typeof options.loggerFn === "function") {
             extra.log = options.loggerFn;
+        } else if (options.loggerFn === false) {
+            /** */ // eslint-disable-line valid-jsdoc
```
JamesHenry: This is a bit odd 😄  Why not put the `// eslint-disable-line valid-jsdoc` on the line you are disabling? Alternatively you could use `Function.prototype` as the value of the noop

j-f1: If I don’t add the JSDoc comment, it errors at line 98 instead. I’ll switch to `Function.prototype`.

JamesHenry:
Thanks a lot for this @j-f1!

PR #354, opened 2017-08-15
Breaking: Only add .implements/.accessibility/.decorators if "truthy"

JamesHenry:
This changes the way `implements`, `accessibility`  and `decorators` are applied.

When running babylon AST comparison tests, I noticed that babylon only applies the above properties if they are "truthy" (not exactly the right way to describe it but hopefully it's clear what I mean).

For example, if there is at least one decorator on a node, there will be a `decorators` array in the AST, otherwise, there will be no `decorators` property at all.

Currently in `typescript-eslint-parser`, we usually apply an empty array for the `decorators` in that case.

Before reviewing and merging this PR, I would first like to confirm that this is indeed the desired behaviour, @andy-ms?

I am personally pretty ambivalent about which makes more sense and would be keen to hear your thoughts on why you opted from leaving them out of the AST vs empty array/null value.

cc @hzoo @soda0289

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -230,6 +230,26 @@ module.exports = function convert(config) {
     }
 
     /**
+     * Converts an array of TSNode parameters into an array of ESTreeNode params
+     * @param  {TSNode[]} parameters An array of TSNode params to be converted
+     * @returns {ESTreeNode[]}       an array of converted ESTreeNode params
+     */
+    function convertParameters(parameters) {
+        if (!parameters || !parameters.length) {
+            return [];
+        }
+        return parameters.map(param => {
+            const convertedParam = convertChild(param);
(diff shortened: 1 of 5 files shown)


andy-ms:
This isn't a big issue for me either; I was following the behavior of the flow plugin, which only adds plugin-specific properties (such as `.typeAnnotation`) if plugin-specific syntax is present. I don't know if this is actually important anywhere in the babel toolchain, but it does make the JSON representation cleaner.

JamesHenry:
Ok, thanks a lot for the feedback! I can't think of any strong reason not to do that (if downstream tools like prettier or the typescript plugin are assuming the properties exist that will be easy enough to fix) I'll request a review from @soda0289, and he can voice any concerns he may have if any.

PR #360, opened 2017-08-18
Chore: Breakout and label TS-specific AST comparison tests

JamesHenry:
Very minor follow up to my last AST comparison PR. This breaks out the typescript fixtures and labels why they currently fail.


soda0289:
Amazing work so far @JamesHenry, thanks for organizing this!

PR #362, opened 2017-08-19
New: Improve enum handling

### LINKED ISSUE #345: `export enum` doesn’t register as an export
2.3.4

4.0.0

```ts
// foo.ts
export enum Foo {
  Bar,
  Baz,
}
// bar.ts
/* eslint "import/named": "error" */
import { Foo } from './foo'
```

No errors — `Foo` is exported in `foo.ts`, and it’s imported in `bar.ts`

**What happened?**

```
Foo not found in './foo.ts' (import/named)
```

---

It looks like this should be parsed as
```
ExportNamedDeclaration {
  // ...
  declaration: TSEnumDeclaration
  // ...
}
```
instead of
```
TSEnumDeclaration {
  // ...
  modifiers: [
    TSExportKeyword
  ]
  // ...
}
```

Also, the `TSEnumDeclaration` node should have the `name` property aliased to `id` so it’s compatible with `ClassDeclaration` nodes.

j-f1:
* Add an `id` property to the generated AST that makes the node more compatible with a `ClassDeclaration` node
* Wrap the `EnumDeclaration` in an `ExportNamedDeclaration` if the enum is exported

Also:

* Show the full stack when a test gets an error about an unknown AST type
* Allow calling `deeplyCopy` with a destination object to copy the data into instead of `result`

Fixes #345.

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -595,6 +595,37 @@ module.exports = function convert(config) {
             });
             break;
 
+        case SyntaxKind.EnumDeclaration:
+            // Sets the enum's `id` property to its `name`
+            // for compatibility with class declarations
+            node.id = node.name;
+            if (node.modifiers && node.modifiers.length) {
+                const mods = node.modifiers;
+                delete node.modifiers;
+                let foundUsefulModifier = false;
+                for (const modifier of mods) {
+                    if (modifier.kind === SyntaxKind.ExportKeyword) {
+                        Object.assign(result, {
+                            type: AST_NODE_TYPES.ExportNamedDeclaration,
+                            declaration: deeplyCopy({
+                                range: [mods.end + 1, node.end],
+                                loc: nodeUtils.getLocFor(mods.end + 1, node.end, ast)
+                            }),
+                            specifiers: [],
+                            source: null
+                        });
+                        foundUsefulModifier = true;
+                        break;
+                    }
+                }
+                if (!foundUsefulModifier) {
+                    deeplyCopy();
+                }
+            } else {
+                deeplyCopy();
+            }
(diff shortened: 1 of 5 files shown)


j-f1:
Since the first commit doesn’t fix an open issue, I can’t add the requested tag to it. I also don’t want to squash the commits because that would make them less clear. What should I do?

JamesHenry:
Thanks a lot for this, @j-f1!

> Allow calling deeplyCopy with a destination object to copy the data into instead of result

How come? You do not use this anywhere in your code

> Breaking-maybe: Don’t copy over properties starting with `_`

I think we should always err on the side of caution and convert this to "breaking", no downside to doing so IMO, but I am not sure this change is necessary for this solution.

Additionally, I do not think we need the changes in node-utils, we should look to explicitly convert nodes where we are handling them, as you can see `deeplyCopy()` is only intended to be used as a fallback, not really as part of an explicit case.

----------

UPDATE: I wrote these comments initially but have since done some digging across this project and babylon...

One really important bit of context for you is that I am currently actively working with the TS Team on aligning on the AST produced by babylon. Therefore any changes we make to AST structure in this project from now on need to take that into account or we'll end up making more changes than we need to. Not everything is black and white and there is still a lot of back and forth to be had, so I apologise if that delays things a bit in any subsequent issues/PRs.

I am also on the Babel team, so I also have babylon source open and have taken your ideas and built on them a little on a separate branch. I have addressed the points I raised above, so please feel free to check it out.

I'm really sorry to have to close this PR in favour of that one. I do really appreciate your work here and it helped bring the enum differences to light really effectively!

Thanks again, @j-f1!

PR #364, opened 2017-08-19
Breaking: Explicitly handle TSEnumDeclaration (fixes #345)

### LINKED ISSUE #345: `export enum` doesn’t register as an export

JamesHenry:
cc @j-f1, as per my comments on the other PR, for this solution I opted for explicitly handling the `TSEnumDeclaration`, and I also wanted to make sure that we (typescript-eslint-parser and babylon) were aligned on the structure of `members`, so I have also handled `TSEnumMember`s and verified this is the case.

I don't believe the other elements from that PR were necessary to fix #345 in the end, so I left them out, but please let me know if you think I am missing something!

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "license": "BSD-2-Clause",
   "devDependencies": {
     "babel-code-frame": "^6.22.0",
-    "babylon": "^7.0.0-beta.19",
+    "babylon": "^7.0.0-beta.20",
     "eslint": "3.19.0",
     "eslint-config-eslint": "4.0.0",
     "eslint-plugin-node": "4.2.2",
--- a/lib/ast-node-types.js
+++ b/lib/ast-node-types.js
@@ -109,6 +109,7 @@ module.exports = {
     TSConstructSignature: "TSConstructSignature",
     TSDeclareKeyword: "TSDeclareKeyword",
     TSEnumDeclaration: "TSEnumDeclaration",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
   "license": "BSD-2-Clause",
   "devDependencies": {
     "babel-code-frame": "^6.22.0",
-    "babylon": "^7.0.0-beta.19",
+    "babylon": "^7.0.0-beta.20",
```
JamesHenry: FYI `7.0.0-beta.20` contained an important fix for exported TS constructs so that is why I have set that as the new baseline

PR #369, opened 2017-08-20
New: Support TypeScript 2.5 (fixes #368)

### LINKED ISSUE #368: Support TypeScript 2.5
https://blogs.msdn.microsoft.com/typescript/2017/08/17/announcing-typescript-2-5-rc/

- [x] Optional Catch Binding

azz:
Only change made was to support optional catch binding:

```ts
try {} catch {}
try {} catch {} finally {}
```

I've done so in the same way that babylon has done, setting `param` to `null` on the `CatchClause`.

Might want to hold out and wait for the full release?

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -497,7 +497,7 @@ module.exports = function convert(config) {
         case SyntaxKind.CatchClause:
             Object.assign(result, {
                 type: AST_NODE_TYPES.CatchClause,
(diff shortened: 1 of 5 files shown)


JamesHenry:
Thanks! I have changed the base to be a new `ts-2.5` branch. We will do what we did last time, keep 2.5 on a protected parallel branch to master until it becomes the stable release

PR #378, opened 2017-08-24
Fix: Location data for typeAnnotations

JamesHenry:
These location data points are obviously not too important to any downstream tools right now, and the changes I have made appear to better align us with `babylon` and the precendents set by `flow`

I have also aligned better with `babylon` as per discussion with Andy in another issue around the default value of many custom modifiers being `undefined`.


JamesHenry:
I added some further detail on remaining differences in the comments in spec.js

PR #388, opened 2017-09-25
Breaking: Updates to AST node types of some TSNodes (fixes #386)

### LINKED ISSUE #386: Breaking Change to lots of AST node types for TypeScript constructs
Follow up to https://github.com/babel/babylon/pull/710

cc @soda0289 @azz @weirdpattern

I should hopefully get to this this week.

JamesHenry:


lib/ast-node-types.js
```diff
@@ -140,17 +139,16 @@ module.exports = {
     TSStaticKeyword: "TSStaticKeyword",
     TSStringKeyword: "TSStringKeyword",
     TSSymbolKeyword: "TSSymbolKeyword",
+    TSTypeAnnotation: "TSTypeAnnotation",
     TSTypeLiteral: "TSTypeLiteral",
+    TSTypeParameter: "TSTypeParameter",
```
existentialism: Looks like this will conflict with `TypeParameter`s that were already being copied over :(

Example with mapped types:

`var x: { [A in keyof B]?: any };`

Ref: https://github.com/prettier/prettier/pull/1393

JamesHenry: Thanks a lot for your patience on this @existentialism, I have a fix for that in prettier and have opened a PR here: https://github.com/prettier/prettier/pull/3455

PR #390, opened 2017-09-30
Fix: Correctly support empty type parameters

j-f1:
This makes the parser handle the `let a: Foo<>` case properly.

Unfortunately, the change seems to cause a regression in one test due to a bug in the TypeScript parser. Any ideas about why this happens would be appreciated.

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -102,17 +102,13 @@ module.exports = function convert(config) {
      * @returns {TypeParameterInstantiation} TypeParameterInstantiation node
      */
     function convertTypeArgumentsToTypeParameters(typeArguments) {
-        const firstTypeArgument = typeArguments[0];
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks so much for your patience on this @j-f1! You both highlighted a fundamental issue in the way we were handling that and got the ball rolling on a fix. Your contributions of the last few months to both the parser and the plugin have been very valuable, so thank you!

I created an issue for it, and included the minimal possible test case from yours as part of the overall fix which matches up with babylon's output: https://github.com/eslint/typescript-eslint-parser/pull/396

I do not make a habit of closing people's PRs in favour of my own, but I am super keen to push on with my greatly improved availability for open source this week, so that is the route I am going to take on this occassion.

Thank you once again!

PR #411, opened 2017-12-10
Chore: Introduce integration tests

JamesHenry:


tests/integration/utils.js
```diff
@@ -0,0 +1,91 @@
...
+        parserOptions: {
+            ecmaVersion: 8,
+            ecmaFeatures: {
+                jsx: true,
+                experimentalObjectRestSpread: true,
+                globalReturn: true
+            },
+            sourceType: sourceType || "script"
+        },
+        cwd: path.join(__dirname)
```
not-an-aardvark: If I'm not mistaken, this could just be `cwd: __dirname`.

tests/integration/utils.js
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    if (overrideConfig) {
+        Object.assign(cliEngineConfig, overrideConfig);
+    }
+
+    const cli = new eslint.CLIEngine(cliEngineConfig);
```
not-an-aardvark: I think it would be possible to just use `eslint.Linter` here:

```js
const linter = new eslint.Linter();

linter.defineParser("typescript-eslint-parser", require("../../parser"));
linter.defineRules(rules);
const messages = linter.verify(
  code,
  {
    parser: 'typescript-eslint-parser',
    parserOptions: {
      ecmaVersion: 8,
      ecmaFeatures: {
        jsx: true,
        experimentalObjectRestSpread: true,
        globalReturn: true
      },
      sourceType: sourceType || "script"
    }
  },
  { filename: 'fake-filename.js' }
)
```

tests/integration/from-babel-eslint.spec.js
```diff
@@ -0,0 +1,1738 @@
+"use strict";
+
+const unpad = require("dedent");
+
+const utils = require("./utils");
+const verifyAndAssertMessages = utils.verifyAndAssertMessages;
+
+describe("Non-regression tests from babel-eslint", () => {
```
not-an-aardvark: Do we need to copy-paste `babel-eslint`'s license here since the tests are from `babel-eslint`?

tests/integration/from-babel-eslint.spec.js
```diff
@@ -0,0 +1,1738 @@
...
+    });
+
+    // describe("flow", () => {
+    //     it("check regular function", () => {
+    //         verifyAndAssertMessages(
+    //             "function a(b, c) { b += 1; c += 1; return b + c; } a;",
+    //             { "no-unused-vars": 1, "no-undef": 1 },
+    //             []
+    //         );
+    //     });
```
not-an-aardvark: Should these commented tests be removed?

tests/integration/from-babel-eslint.spec.js
JamesHenry: Yes, definitely

tests/integration/utils.js
JamesHenry: Ah yeah, that used to be nested that's why there is the join there!

tests/integration/utils.js
JamesHenry: Unless I am mistaken when using the Linter, it is not possible to use `extends`. Being able to run integration tests concisely using presets is important IMO.

not-an-aardvark: You're correct that it's not possible to use `extends` with Linter, but it doesn't seem like `extends` is being used for any of these tests. Were you planning to add it?

(I don't mean to block the PR with this concern -- I'm just bringing it up in case you made a mistake.)

tests/integration/from-babel-eslint.spec.js
JamesHenry: Done

tests/integration/utils.js
JamesHenry: The one I pasted in directly above, line 20?

JamesHenry:
I have cleaned up the babel-eslint tests somewhat (all the flow ones now gone, for example), but have left in some commented out tests as they will form a useful basis for the custom Referencer which will come in a follow up PR.

tests/integration/utils.js
not-an-aardvark: Oops, I missed that, sorry. (I hadn't realized that this function was also being used from other files.)

JamesHenry: Yeah I wasn't a huge fan of the huge single file used by babel-eslint so I extracted it and refactored it within a helper function.

Thanks for the review!

Issue #402, opened 2017-10-26
no-redeclare throws errors on overloaded functions


j-f1:
**What version of TypeScript are you using?**

2.5.3

**What version of `typescript-eslint-parser` are you using?**

8.0.0

**What code were you trying to parse?**

```ts
export async function readFile(
  filename: string,
  options?: { flag?: string }
): Promise<Buffer>
export async function readFile(
  filename: string,
  options?: { encoding: BufferEncoding; flag?: string }
): Promise<string>
export async function readFile(
  filename: string,
  options?: { encoding?: string; flag?: string }
): Promise<Buffer | string> {
  // ...
}
```

**What did you expect to happen?**

No errors — `readFile` is only declared once.

**What happened?**

```
'readFile' is already defined.
```

JamesHenry:
We should hopefully be able to solve this with upcoming custom Referencer work

JamesHenry:
This was finally fixed by https://github.com/eslint/typescript-eslint-parser/pull/412. I added the source above as an integration test.

Issue #389, opened 2017-09-26
Overloaded method signatures failing on certain eslint rules


ismail-syed:
**What version of TypeScript are you using?**
`2.5.2`

**What version of `typescript-eslint-parser` are you using?**
`8.0.0`

**What code were you trying to parse?**

```
export class Test {
  constructor() {
  }
  public test(param1: Number): Test;
  public test(param1: Test): Test;
  public test(param1: any): Test {
      return new Test();
    }
}
```

**What did you expect to happen?**
With the following rules enabled, I would expect the code above to be parsed. The issue seems to be that parsing overloaded method signatures, with the following eslint rules enabled fails. 

```
    "rules": {
      "array-callback-return": "error",
      "getter-return": "error",
      "strict": "error",
      "lines-around-directive": "error",
      "no-empty-function": "error"
    }
```

Turning off all those rules, and enabling them individually fails as well.  

**What happened?**

```
$ eslint . --ext .ts --max-warnings 0
Cannot read property 'type' of null
TypeError: Cannot read property 'type' of null
    at EventEmitter.onCodePathStart (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/eslint/lib/rules/array-callback-return.js:193:34)
n.js:193:34)
    at emitTwo (events.js:130:20)
    at EventEmitter.emit (events.js:213:7)                                                                                                                           lyzer.js:360:30)
    at processCodePathToEnter (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/eslint/lib/code-path-analysis/code-path-ana-analyzer.js:603:9)
lyzer.js:360:30)
    at CodePathAnalyzer.enterNode (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/eslint/lib/code-path-analysis/code-path
-analyzer.js:603:9)
    at Traverser.enter (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/eslint/lib/linter.js:962:32)    at Traverser.__execute (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/estraverse/estraverse.js:397:31)
    at Traverser.traverse (/Users/ismailsyed/src/github.com/ismail-syed/typescript-eslint-parser-example/node_modules/estraverse/estraverse.js:501:28)
[...]

PR #412, opened 2017-12-12
Breaking: Implement parseForESLint() function

### LINKED ISSUE #400: array-callback-return TypeError
2.5.3

8.0.1

```javascript
export abstract class TestAbstract {

  onCreated?(): Promise<void> | any;

}
```

ESLint shouldnt fail when parsing the file.

**What happened?**
`TypeError: Cannot read property 'type' of null`
`node_modules/eslint/lib/rules/array-callback-return.js:194:34`

If I add a function body, it works:
```javascript
  onCreated(): Promise<void> | any {

  }
```

JamesHenry:
This PR implements a parseForESLint() function.

This will be preferred by ESLint, and so we can use it to implement tweaks to the AST which are only required in an ESLint context. The idea is that this will be used only where absolutely necessary.

The most pressing case is that of a method's value (`FunctionExpression`) having no body in some valid situations in TypeScript.

For now, the simplest solution is just to namespace the node.type as `TSEmptyBody${node.type}` only in the case that we are parsing for ESLint.

Prettier's usage will be completely unaffected because it will continue to use `parse()`.

Fixes #400 #389 #402 #253 

🎉 

Other minor notes:
- I updated the README, there were some out of date issues in there
- I removed old `babel-eslint` issue numbers from that spec file as they were confusing
- Removed the erroneously committed `yarn.lock` file


j-f1:
Are you sure `TSEmptyBodyFoo` is the best solution? IMHO the `FunctionDeclaration` variant should be `TSFunctionOverloadDeclaration`, and the `MethodDeclaration` variant should be `TSMethodOverloadDeclaration`, since that is what they are used for.

JamesHenry:
@j-f1 I was personally just thinking how the name is very unimportant at this stage. The only stakeholders that could possibly care about these particular node names are custom ESLint TypeScript rule authors (I know you are one of them :D) and all of that goes through `eslint-plugin-typescript` right now. Any non-ESLint users of this parser won't even come across those nodes.

Having a more situational name like `TSFunctionOverloadDeclaration`, describing what it does rather than what it is, would simply require quite a bit more code to cover all the cases.

This seemed like the more straightforward starting point, and we could certainly change it in the future if it makes sense to.

mysticatea:
In my understand, this PR adds 2 node types:

- TSEmptyBodyFunctionExpression
- TSEmptyBodyFunctionDeclaration

Looks good to me.

By the way, I think that some cases should be throw syntax errors. For example:

```js
const obj = {
    a()
}
```

- TypeScript playground shows a syntax error: https://www.typescriptlang.org/play/index.html#src=const%20obj%20%3D%20%7B%0D%0A%20%20%20%20a()%0D%0A%7D
- But `typescript-eslint-parser` does not realize this syntax error: http://astexplorer.net/#/gist/e2dad04ac8feee4f9e19e53872aeef2b/e28c438e3eec65e47d9d6a7be45627b85f77aa76

JamesHenry:
Yes, this is the case in general with the TypeScript compiler (which powers this parser). It is hyper lenient during the parsing phase because it has language services which come later in the pipeline to provide feedback to developers whilst they author TypeScript.

Adding our own strictness on top of that would need to come later in a separate PR, but it might be worth doing at some point.

j-f1:
> Having a more situational name like `TSFunctionOverloadDeclaration`, describing what it does rather than what it is, would simply require quite a bit more code to cover all the cases.

In what other cases would you get one of these bodyless functions?

JamesHenry:
E.g.

```ts
interface I {
    func(): string;
}

declare function myFunc(): boolean;

abstract class C {
    abstract foo(): number;
}
```

tests/integration/from-babel-eslint.spec.js
```diff
@@ -553,18 +553,18 @@ describe("Non-regression tests from babel-eslint", () => {
...
-            { "no-implicit-globals": 1 },
-            [],
-            "module",
-            {
-                env: {},
-                parserOptions: { ecmaVersion: 6, sourceType: "module" }
-            }
-        );
-    });
+    // it("no-implicit-globals in module", () => {
```
kaicataldo: Just for my own knowledge, why are these tests commented out?

j-f1:
@JamesHenry 

```ts
interface I {
    func(): string; // TSMethodSignature
}

declare function myFunc(): boolean; // DeclareFunction

abstract class C {
    abstract foo(): number; // TSAbstractMethodDeclaration with FunctionExpression value
}
```

---

How about `TSMethodAnnotation` and `TSFunctionAnnotation` as types instead?

parser.js
```diff
@@ -177,6 +185,12 @@ exports.version = require("./package.json").version;
 
 exports.parse = parse;
 
+exports.parseForESLint = function parseForESLint(code, options) {
+    options.parseForESLint = true;
```
not-an-aardvark: I think it would be best to avoid mutating the options object provided by the user. (This was also a problem in https://github.com/eslint/espree/issues/329.)

tests/integration/from-babel-eslint.spec.js
JamesHenry: I haven't solved them yet. They need to be solved by a custom Resolver I think, but I didn't want to lost them when I cut the custom resolver out of this PR in favour of this new flag.

It would be fair enough to request I take them out of the PR, but laziness made me want to keep them in as reminders 😄

parser.js
JamesHenry: In which case I would be forced to change the signature of `parse()` to accept an optional third argument. There's no reason that would be an issue, right?

not-an-aardvark: I don't think it would be an issue.

If you don't want the change to be visible to the public API, you could create a helper function, e.g. `parseMaybeForESlint(code, options, internalOptions)`, and have `parse` call `parseMaybeForESLint(code, options, { isForESLint: false })` and have `parseForESLint` call `parseMaybeForESLint(code, options, { isForESLint: true })`.

platinumazure: What about just defensively cloning the options so the user can't see the mutation?

j-f1: You could also do `options = Object.assign({}, options, { parseForESLint: true })`. Or insert `options = Object.create(options)` before this line.

not-an-aardvark: It seems strange to have `parseForESLint` be mixed with the user-provided options object. For example, if the user added `parseForESLint: true` to their options object for some reason, they could cause strange behavior.

JamesHenry:
@not-an-aardvark PTAL at my latest commit https://github.com/eslint/typescript-eslint-parser/pull/412/commits/b273fca52b1f4afc8db6e81491304531fb55b44a

JamesHenry:
@j-f1 `eslint-plugin-typescript` is the only possible stakeholder in those node types right now, so we can discuss again if it becomes an issue with that, and they can easily be changed. I really wanted to get this in ASAP and feel `TSEmptyBodyFunctionExpression` `TSEmptyBodyFunctionDeclaration` are at least very clear as a starting point, given they are replacing `FunctionExpression` and `FunctionDeclaration`.

PR #415, opened 2017-12-16
New: Add some parserServices

j-f1:
This lets you get the `ts.Program` instance for a file, as well as the TypeScript AST node corresponding to a particular node, and the type of any node.

### DIFF
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -57,10 +57,9 @@ module.exports = function convert(config) {
         range: [node.getStart(), node.end],
         loc: nodeUtils.getLoc(node, ast)
     };
-    Object.defineProperty(result, nodeUtils.originalNodeSymbol, {
-        value: node,
-        enumerable: false
-    });
+
+    nodeUtils.map.set(result, node);
+    nodeUtils.reverseMap.set(node, result);
 
     /**
      * Copies the result object into an ESTree node with just a type property.
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -55,7 +55,8 @@ module.exports = function convert(config) {
     let result = {
         type: "",
         range: [node.getStart(), node.end],
-        loc: nodeUtils.getLoc(node, ast)
+        loc: nodeUtils.getLoc(node, ast),
+        [nodeUtils.originalNodeSymbol]: node
     };
 
     /**
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -55,9 +55,12 @@ module.exports = function convert(config) {
     let result = {
         type: "",
         range: [node.getStart(), node.end],
-        loc: nodeUtils.getLoc(node, ast),
-        [nodeUtils.originalNodeSymbol]: node
+        loc: nodeUtils.getLoc(node, ast)
     };
+    Object.defineProperty(result, nodeUtils.originalNodeSymbol, {
+        value: node,
+        enumerable: false
+    });
 
     /**
      * Copies the result object into an ESTree node with just a type property.
--- a/lib/node-utils.js
+++ b/lib/node-utils.js
@@ -98,7 +98,11 @@ TOKEN_TO_TEXT[SyntaxKind.CaretEqualsToken] = "^=";
 TOKEN_TO_TEXT[SyntaxKind.AtToken] = "@";
 TOKEN_TO_TEXT[SyntaxKind.InKeyword] = "in";
 
-const originalNodeSymbol = Symbol("Original TypeScript AST node");
+// ASTNode => TSNode
+const map = new WeakMap();
+
+// TSNode => ASTNode
+const reverseMap = new WeakMap();
 
 /**
  * Find the first matching child based on the given sourceFile and predicate function.
@@ -157,8 +161,8 @@ module.exports = {
      * Expose the enum of possible TSNode `kind`s.
      */
     SyntaxKind,
-    originalNodeSymbol,
-    getTSNode,
+    map,
+    reverseMap,
     isAssignmentOperator,
     isLogicalOperator,
     getTextForTokenKind,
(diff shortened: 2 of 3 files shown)


JamesHenry:
I have a branch from a year ago(!) which was looking to address this: https://github.com/eslint/typescript-eslint-parser/compare/parser-services

I think we should use that Map based implementation rather than adding things to every AST node, and I don't think it is necessary for the parser to concern itself with the type checker yet. Exposing the ts.Program gives plugin authors everything they need right? I think we should look to add the bare minimum that is useful (as this effectively a new public API we are supporting) and then keep an eye on the actual usage. WDYT?

j-f1:
> Exposing the ts.Program gives plugin authors everything they need right?

As seen above in my PR to desktop/desktop, the current usage doesn’t properly handle types imported from other files, or from `globals.d.ts` files. I’m working on a better system, but it needs a patch in ESLint core to send the path of the file being parsed to the parser.

> I think we should use that Map based implementation rather than adding things to every AST node.

👍 Will take a look at using `Map`/`WeakMap`.

j-f1:
`WeakMap` added. Sorry for the delay.

parser.js
```diff
@@ -186,12 +210,15 @@ function generateAST(code, options, additionalParsingContext) {
 exports.version = require("./package.json").version;
 
 exports.parse = function parse(code, options) {
-    return generateAST(code, options, { isParseForESLint: false });
```
JamesHenry: The name `generateAST` no longer makes sense for this function after this change

lib/convert.js
```diff
@@ -58,6 +58,9 @@ module.exports = function convert(config) {
         loc: nodeUtils.getLoc(node, ast)
     };
 
+    nodeUtils.map.set(result, node);
```
JamesHenry: I think these should come right before we return the result, that way you don't need to have two lots of them (i.e. there is no need to have them again in node-utils)

JamesHenry: `map` and `reverseMap` are not clear names IMO

lib/node-utils.js
```diff
@@ -98,6 +98,12 @@ TOKEN_TO_TEXT[SyntaxKind.CaretEqualsToken] = "^=";
 TOKEN_TO_TEXT[SyntaxKind.AtToken] = "@";
 TOKEN_TO_TEXT[SyntaxKind.InKeyword] = "in";
 
+// ASTNode => TSNode
```
JamesHenry: We refer to them as `ESTreeNode` throughout the codebase so please go for that

lib/node-utils.js
```diff
@@ -553,6 +561,9 @@ function fixExports(node, result, ast) {
             loc: getLocFor(exportKeyword.getStart(), result.range[1], ast)
         };
 
+        map.set(newResult, node);
```
JamesHenry: As per comment above I don't think these Maps need to be in node-utils at all

parser.js
```diff
@@ -176,7 +179,28 @@ function generateAST(code, options, additionalParsingContext) {
...
+ * Generate the `parserServices` object for a parse result
+ * @param   {Result} result The return value of `generateAST`
+ * @returns {Object}        The `parserServices` object
+ */
+function getServices(result) {
+    const typeChecker = result.program.getTypeChecker();
+    return {
+        program: result.program,
+        getTSNode: node => nodeUtils.map.get(node),
+        getESNode: tsNode => nodeUtils.reverseMap.get(tsNode),
```
JamesHenry: Again, `ESTreeNode` is the name that has been used.

I'm a little unsure about the use-case for `getESNode` and `reverseMap` as they are currently called. I feel which should err on adding as little as possible to this API surface, particularly in this initial PR, we can always add more later

parser.js
```diff
@@ -176,7 +179,28 @@ function generateAST(code, options, additionalParsingContext) {
...
+        typeChecker,
```
JamesHenry: As per my comment above about minimal API surface - why do we need a reference to the typeChecker, if we are already exposing the created Program?

parser.js
```diff
@@ -176,7 +179,28 @@ function generateAST(code, options, additionalParsingContext) {
...
+        getType(node) {
```
JamesHenry: Helper functions like this could always be added later if we found that the demand is high, but again, we have already exposed everything you would need to get this yourself

JamesHenry:
@j-f1 I am going to try and get a TS 2.7 release out ASAP, but do you think we could pick this up again after that. No worries, if you don't have any availability right now, just let me know.

j-f1:
I gave up on the rule I was writing this for, partially because there were some issues (I don’t remember exactly what) that prevented me from getting accurate type information.

lib/convert.js
j-f1: I set them when the nodes are created so that in e.g.
```ts
export class Foo {}
```
the `ClassDeclaration` and `ExportNamedDeclaration` have a TS AST node attached to them (`ClassDeclaration`).

parser.js
j-f1: I found it really useful when I was writing a rule using this — otherwise you have to do
```js
context.parserServices.program.getTypeChecker().getTypeAtLocation(context.parserServices.getTSNode(node))
```
just to get the type of a node. It’s also arguably the most common use of the parser services.

j-f1:
I feel like the best way to implement this would be to use TS’s language service API.

j-f1:
I’m not working on the project I was going to use this for right now so it’s pretty far down my priority list, and I don’t think I’ll be able to work on this in the near future. Feel free to work off my branch though!

kaicataldo:
Thanks for the update! Should we just close this for now and revisit this issue if it’s ever needed again?

j-f1:
Sounds good 👍

uniqueiniquity:
@j-f1 @JamesHenry Do you know if anyone has been working on this recently? I'm on the TypeScript team, and we'd love to see if there's a way to get semantic information available for rules. If it's not already being tackled, I'd love to pick up this effort and see where we can get with it!

j-f1:
@uniqueiniquity Go ahead! It’d be really cool to see this implemented.

PR #423, opened 2017-12-24
Fix: Remove all tokens inside comments from tokens array (fixes #422)

### LINKED ISSUE #422: Core indent rule reports false positives in doc blocks
`typescript@2.6.2`

`typescript-eslint-parser@11.0.0`

```ts
/**
 * This is a function.
 * @param {String} text
 * @returns {String}
 */
function foo(text: string): string {}

```

The minimal repro case is:
```ts
/**
 * This is a function.
 * @param {String} text
 * @returns {String}
 */
function foo() {}
```

Running the autofixer generates the following code (which then passes):
```ts
 /**
 * This is a function.
{String} text
{String}
 */
function foo(text: string): string {}
```

No errors! This only occurs with the typescript-eslint-parser. It does not happen with the default parser.

**What happened?**
```sh
/Users/cataldo/Code/test/eslint-test/index.js
  3:1  error  Expected indentation of 0 spaces but found 3  indent
  4:1  error  Expected indentation of 0 spaces but found 3  indent
```

I'll try to look into the cause if I get a chance, but not sure if and when that'll be. Took the bug label off because I'm not sure what the established practice is in this repo - happy to update though :)

kaicataldo:
This fixes an issue where TypeScript generates tokens for `{Type}` (for both the brackets and the Identifier `Type`), causing the indent rule in core to have some false positives. I'm assuming these tokens are generated as part of TypeScript's JS file checking and type hinting from doc blocks.

@eslint/eslint-team As an aside, could we potentially leverage this behavior for our own JSDoc parsing and validation in ESLint core? Would be nice if we could rely on a well-maintained project like TypeScript instead of doctrine, since we've been struggling to maintain that.

Maybe @JamesHenry has some idea of the feasibility of this, given your experience!

This is my first time working with this codebase or really anything TypeScript related, so please do let me know if there are better ways to achieve this or if any assumptions I've made are incorrect.

To illustrate the change, here's a before and after of the tokens and comments arrays generated for the following code:

```js
/**
 * This is a function.
 * @param {String} text
 * @returns {String}
 */
function foo(text) {}
```
Before this patch:
```
Tokens:  [ { type: 'Punctuator',
    value: '{',
    range: [ 37, 38 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'String',
    range: [ 38, 44 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '}',
    range: [ 44, 45 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'text',
    range: [ 46, 50 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '{',
    range: [ 63, 64 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'String',
    range: [ 64, 70 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '}',
    range: [ 70, 71 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Keyword',
    value: 'function',
    range: [ 76, 84 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'foo',
    range: [ 85, 88 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '(',
    range: [ 88, 89 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'text',
    range: [ 89, 93 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: ')',
    range: [ 93, 94 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '{',
    range: [ 95, 96 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '}',
    range: [ 96, 97 ],
    loc: { start: [Object], end: [Object] } } ]

Comments:  [ { type: 'Block',
    value: '*\n * This is a function.\n * @param {String} text\n * @returns {String}\n ',
    range: [ 0, 75 ],
    loc: { start: [Object], end: [Object] } } ]
```

After this patch:
```
Tokens:  [ { type: 'Keyword',
    value: 'function',
    range: [ 76, 84 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'foo',
    range: [ 85, 88 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '(',
    range: [ 88, 89 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Identifier',
    value: 'text',
    range: [ 89, 93 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: ')',
    range: [ 93, 94 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '{',
    range: [ 95, 96 ],
    loc: { start: [Object], end: [Object] } },
  { type: 'Punctuator',
    value: '}',
    range: [ 96, 97 ],
    loc: { start: [Object], end: [Object] } } ]

Comments:  [ { type: 'Block',
    value: '*\n * This is a function.\n * @param {String} text\n * @returns {String}\n ',
    range: [ 0, 75 ],
    loc: { start: [Object], end: [Object] } } ]
```

### DIFF
--- a/lib/node-utils.js
+++ b/lib/node-utils.js
@@ -245,6 +245,32 @@ function isComma(token) {
 }
 
 /**
+ * Returns true if the given TSToken is a comment
+ * @param  {TSToken}  token the TypeScript token
+ * @returns {boolean}       is commment
+ */
+function isComment(token) {
+    return (token.kind === SyntaxKind.SingleLineCommentTrivia || token.kind === SyntaxKind.MultiLineCommentTrivia) || (token.kind >= SyntaxKind.JSDocTypeExpression && token.kind <= SyntaxKind.JSDocTypeLiteral);
+}
+
+/**
+ * Returns true if the given TSToken is inside a comment
+ * Non-comment type tokens are generated for types in JSDoc Blocks
+ * @param  {TSToken}  token the TypeScript token
+ * @returns {boolean}       is inside a commment
+ */
+function isInsideComment(token) {
+    if (token.kind === SyntaxKind.SourceFile) {
+        return false;
+    }
+    if (isComment(token)) {
+        return true;
+    }
+    return isInsideComment(token.parent);
+}
+
+
+/**
  * Returns the binary expression type of the given TSToken
  * @param  {TSToken} operator the operator token
  * @returns {string}          the binary expression type
@@ -693,7 +719,7 @@ function convertTokens(ast) {
      * @returns {undefined}
      */
     function walk(node) {
-        if (isToken(node) && node.kind !== SyntaxKind.EndOfFileToken) {
+        if (isToken(node) && !isInsideComment(node) && node.kind !== SyntaxKind.EndOfFileToken) {
             const converted = convertToken(node, ast);
 
             if (converted) {
--- a/lib/node-utils.js
+++ b/lib/node-utils.js
@@ -185,7 +185,9 @@ module.exports = {
     convertTokens,
     getNodeContainer,
     isWithinTypeAnnotation,
-    isTypeKeyword
+    isTypeKeyword,
+    isComment,
+    isJSDocComment
 };
 /* eslint-enable no-use-before-define */
 
@@ -245,30 +247,22 @@ function isComma(token) {
 }
 
 /**
- * Returns true if the given TSToken is a comment
- * @param  {TSToken}  token the TypeScript token
- * @returns {boolean}       is commment
+ * Returns true if the given TSNode is a comment
+ * @param {TSNode} node the TypeScript node
+ * @returns {boolean} is commment
  */
-function isComment(token) {
-    return (token.kind === SyntaxKind.SingleLineCommentTrivia || token.kind === SyntaxKind.MultiLineCommentTrivia) || (token.kind >= SyntaxKind.JSDocTypeExpression && token.kind <= SyntaxKind.JSDocTypeLiteral);
+function isComment(node) {
+    return node.kind === SyntaxKind.SingleLineCommentTrivia || node.kind === SyntaxKind.MultiLineCommentTrivia;
 }
 
 /**
- * Returns true if the given TSToken is inside a comment
- * Non-comment type tokens are generated for types in JSDoc Blocks
- * @param  {TSToken}  token the TypeScript token
- * @returns {boolean}       is inside a commment
+ * Returns true if the given TSNode is a JSDoc comment
+ * @param {TSNode} node the TypeScript node
+ * @returns {boolean} is JSDoc comment
  */
-function isInsideComment(token) {
-    if (token.kind === SyntaxKind.SourceFile) {
-        return false;
-    }
-    if (isComment(token)) {
-        return true;
-    }
-    return isInsideComment(token.parent);
+function isJSDocComment(node) {
+    return node.kind === SyntaxKind.JSDocComment;
 }
-
 
 /**
  * Returns the binary expression type of the given TSToken
@@ -282,7 +276,6 @@ function getBinaryExpressionType(operator) {
         return "LogicalExpression";
     }
     return "BinaryExpression";
-
 }
 
 /**
@@ -719,7 +712,13 @@ function convertTokens(ast) {
      * @returns {undefined}
      */
     function walk(node) {
-        if (isToken(node) && !isInsideComment(node) && node.kind !== SyntaxKind.EndOfFileToken) {
+        // TypeScript generates tokens for types in JSDoc blocks. Comment tokens
+        // and their children should not be walked or added to the resulting tokens list.
+        if (isComment(node) || isJSDocComment(node)) {
+            return;
+        }
+
+        if (isToken(node) && node.kind !== SyntaxKind.EndOfFileToken) {
             const converted = convertToken(node, ast);
 
             if (converted) {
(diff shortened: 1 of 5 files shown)


j-f1:
Looks good!

Can you delete `package-lock.json.1799798656`?

lib/node-utils.js
```diff
@@ -244,6 +244,32 @@ function isComma(token) {
...
+}
+
+/**
+ * Returns true if the given TSToken is inside a comment
+ * Non-comment type tokens are generated for types in JSDoc Blocks
+ * @param  {TSToken}  token the TypeScript token
+ * @returns {boolean}       is inside a commment
+ */
+function isInsideComment(token) {
+    if (token.kind === SyntaxKind.SourceFile) {
```
kaicataldo: Is this check sufficient? I was wondering if the algorithm could alternatively be:

kaicataldo: We could also avoid crawling up the tree on every node if we knew exactly what kind of tokens can exist in JSDoc comments (scoping the check to those node types). Maybe someone with more knowledge of could help with that?

lib/node-utils.js
```diff
@@ -693,7 +719,7 @@ function convertTokens(ast) {
      * @returns {undefined}
      */
     function walk(node) {
```
kaicataldo: I also thought about keeping somehow keeping track of if the parent node is a comment and passing it through the recursive `walk` function here (it could just be a boolean flag or options object). Now that I think about it, that might actually be the best way to achieve this since we probably don't really want to be potentially crawling up the tree on every node when we don't have to.

And again, if there's a better way to tell if a node is inside a comment from some value on the AST or some util we already have, please let me know!

kaicataldo:
@j-f1 Whoops! Not sure how that got in there. Thanks.

lib/node-utils.js
```diff
@@ -244,6 +244,15 @@ function isComma(token) {
     return token.kind === SyntaxKind.CommaToken;
 }
 
+/**
+ * Returns true if the given TSToken is a comment
+ * @param  {TSToken}  token the TypeScript token
+ * @returns {boolean}       is commment
+ */
+function isComment(token) {
+    return (token.kind === SyntaxKind.SingleLineCommentTrivia || token.kind === SyntaxKind.MultiLineCommentTrivia) || (token.kind >= SyntaxKind.JSDocTypeExpression && token.kind <= SyntaxKind.JSDocTypeLiteral);
```
kaicataldo: Is this the correct check for comment tokens?

kaicataldo:
I realized this could be simplified immensely. Since we're walking the tree from the top down, we can simply stop walking whenever we encounter a comment. This way, it's children will never be converted or added to the token list.

lib/node-utils.js
kaicataldo: Realized there's an even simpler way to do this - please see later commits.

lib/node-utils.js
```diff
@@ -244,6 +244,15 @@ function isComma(token) {
     return token.kind === SyntaxKind.CommaToken;
 }
 
+/**
+ * Returns true if the given TSToken is a comment
+ * @param  {TSToken}  token the TypeScript token
+ * @returns {boolean}       is commment
+ */
+function isComment(token) {
+    return token.kind === SyntaxKind.SingleLineCommentTrivia || token.kind === SyntaxKind.MultiLineCommentTrivia || (token.kind >= SyntaxKind.JSDocTypeExpression && token.kind <= SyntaxKind.JSDocTypeLiteral);
```
kaicataldo: Is this the correct check for comment tokens?

kaicataldo:
Friendly ping @JamesHenry @soda0289

lib/node-utils.js
JamesHenry: TypeScript themselves use this within their private utilities:

```ts
function isComment(kind: SyntaxKind): boolean {
  return kind === SyntaxKind.SingleLineCommentTrivia || kind === SyntaxKind.MultiLineCommentTrivia;
}
```

I feel like we should split the JSDoc checks out into a separate function.

The TypeScript team has these (again not in their public API, but useful utils):

```ts
// JSDoc
    /** True if node is of some JSDoc syntax kind. */
    /* @internal */
    export function isJSDocNode(node: Node): boolean {
        return node.kind >= SyntaxKind.FirstJSDocNode && node.kind <= SyntaxKind.LastJSDocNode;
    }

    /** True if node is of a kind that may contain comment text. */
    export function isJSDocCommentContainingNode(node: Node): boolean {
        return node.kind === SyntaxKind.JSDocComment || isJSDocTag(node);
    }

    // TODO: determine what this does before making it public.
    /* @internal */
    export function isJSDocTag(node: Node): boolean {
        return node.kind >= SyntaxKind.FirstJSDocTagNode && node.kind <= SyntaxKind.LastJSDocTagNode;
    }
```

kaicataldo:
Updated! Thanks for the review :)

kaicataldo:
Friendly ping!

JamesHenry:
I think this is great for now, thanks! I'm going to change it to be a `Breaking` change though just in case anybody was depending on the old behaviour somehow

kaicataldo:
Thanks!

PR #429, opened 2018-01-09
Fix: Remove "Experimental" from rest and spread (fixes #428)

### LINKED ISSUE #428: Rename ExperimentalSpreadProperty

2.6.2

master

```ts
const foo = { ...bar };
```

An `SpreadElement` or `SpreadProperty` should be in the AST, not `ExperimentalSpreadProperty`, since they are not considered experimental anymore (https://github.com/Microsoft/TypeScript/commit/39b9163715fac6975c5691ae4d85afbc2eabcb62)?

I'm not exactly sure how this should go but would make things a lot easier to maintain in prettier (for example https://github.com/prettier/prettier/pull/3681) :-)

I can help with this, just need a direction (cc @JamesHenry)

duailibe:
I guess this is a breaking change..

cc @JamesHenry

### DIFF
--- a/lib/ast-node-types.js
+++ b/lib/ast-node-types.js
@@ -42,8 +42,6 @@ module.exports = {
     Decorator: "Decorator",
     DoWhileStatement: "DoWhileStatement",
     EmptyStatement: "EmptyStatement",
-    ExperimentalRestProperty: "ExperimentalRestProperty",
-    ExperimentalSpreadProperty: "ExperimentalSpreadProperty",
     ExportAllDeclaration: "ExportAllDeclaration",
     ExportDefaultDeclaration: "ExportDefaultDeclaration",
     ExportNamedDeclaration: "ExportNamedDeclaration",
--- a/lib/convert.js
+++ b/lib/convert.js
@@ -1133,10 +1133,8 @@ module.exports = function convert(config) {
 
                 if (node.dotDotDotToken) {
                     Object.assign(result, {
-                        type: AST_NODE_TYPES.ExperimentalRestProperty,
-                        argument: convertChild(node.propertyName || node.name),
-                        computed: Boolean(node.propertyName && node.propertyName.kind === SyntaxKind.ComputedPropertyName),
-                        shorthand: !node.propertyName
+                        type: AST_NODE_TYPES.RestElement,
+                        argument: convertChild(node.propertyName || node.name)
                     });
                 } else {
                     Object.assign(result, {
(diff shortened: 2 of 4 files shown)


j-f1:
I think this is breaking since the experimental nodes won’t be present in the AST anymore.

kaicataldo:
I'm not sure we can do this. ESLint itself still considers these experimental and so refers to these node types as such in all of its core rules. We'll most likely change that in ESLint core in the next major release after these hit stage 4 (currently [stage 3](https://github.com/tc39/proposals#stage-3)), but this will require changing all the core rules to account for the new node type as well.

JamesHenry:
Thanks @duailibe, yes I raised this myself a few months ago. @kaicataldo's feedback is the same conclusion that was drawn then.

However, one interesting thing has changed since then. Not too long ago I introduced the `parseForESLint` vs `parse` distinction.

Prettier uses this parser via `parse`, ESLint ecoystem via `parseForESLint`, and we already have one existing case where we deviate on AST ( see `namespaceEmptyBodyFunctionForESLint()`).

If you are up for it @duailibe, it might be worth adding a new helper to allow for us to produce this newer AST from `parse` and stick with the old one from `parseForESLint`.

I obviously want to keep deviations on AST between the two functions to an absolute minimum, but a case such as this (where ESLint plans to change anyway), might be worth considering.

What do you think @kaicataldo?

kaicataldo:
I would also lean towards minimizing differences, but if changing this for `parse` would improve things for other downstream tools, I'm not against that!

Once this syntax reaches stage 4 these should converge again anyway :)

duailibe:
@JamesHenry @kaicataldo thank you both!

That seems like a fair compromise, I'll work on that direction sometime today or this weekend.

JamesHenry:
@kaicataldo This is now stage-4: https://github.com/tc39/proposals/blob/a41dccc91675f41c65dee354c18c19946d6f2f45/finished-proposals.md

Shall we just go ahead and forget about the interim solution and use this PR?

mysticatea:
https://github.com/eslint/eslint/pull/9943 has been merged. We can use `RestElement`/`SpreadElement` for rest/spread properties since the next version of ESLint.

JamesHenry:
Thanks, @mysticatea!

JamesHenry:
We could also potentially leverage the (now deprecated) `ecmaFeatures.experimentalObjectRestSpread flag` to allow users to opt into the old behaviour when using older versions of ESLint...

kaicataldo:
Now that we're close to releasing ESLint v5, I wonder if we can just merge this as a breaking change? Otherwise, the suggestion made [here](https://github.com/eslint/typescript-eslint-parser/pull/429#issuecomment-366773892) sounds good!

JamesHenry:
Thank you again, @duailibe, and sorry for how long this has taken to go through!

I think we should merge this as a breaking change.

Some important changes went in which affected the `ecma-features` snapshot, so I'm afraid this PR now has conflicts which have to be resolved locally.

@duailibe Will you have chance to do that soon? No worries if not, I can finish it off

duailibe:
@JamesHenry don't worry! I can resolve the conflicts, and will update the PR in a bit.

JamesHenry:
Awesome, thanks @duailibe!

PR #436, opened 2018-01-22
Breaking: Infer parsing as JSX from file extension (fixes #399)

### LINKED ISSUE #399: Investigate the possibility of setting `jsx: true` based on file extension
In this parser we have to use the `parserOptions.jsx` flag to determine whether or not to configure the actual TypeScript compiler to parse the source as JSX.

When the TypeScript compiler parses certain syntax in "JSX mode" it is ambiguous and could be a JSXElement as far as it's concerned.

This is the same reason you cannot use the `<>` syntax for type assertions in TSX files, you have to use the `as` syntax.

Essentially then, if you want to have `jsx: true`, you have to accept that you are always in "JSX mode" in TypeScript, working in the same way as if all of your source files were .tsx, and so you cannot use any ambiguous constructs.

Right now the DX is not at all clear when you have configured `jsx: true` in your ESLint config and then use ambiguous TypeScript constructs. You will just end up with a random parse error from the TypeScript compiler.

Ideally, we would be able to provide clearer error messaging, or infer the jsx setting itself from the file extension (.ts vs .tsx)

This would have avoided the need for users to report:

https://github.com/eslint/typescript-eslint-parser/issues/392
https://github.com/eslint/typescript-eslint-parser/issues/393

kaicataldo:
fixes #399 

This is a WIP implementation of inferring whether the parser should parse in JSX mode or not based on the file extension of the file being parsed. Wanted to have a place to start a discussion, and then when we decide how we want to proceed finish this up.

This is a breaking change for users who are running ESLint on files (and therefore have a filepath). However, I can't think of a reason why you'd ever want the behavior to be different.

That being said, I believe that we do need to keep the behavior to set it manually for when a filepath is not provided, since we don't know whether it should be parsed in JSX mode or not. Example:
```sh
echo 'var foo: number = 1;' | ./node_modules/.bin/eslint --stdin --no-eslintrc --parser-options='ecmaFeatures:{jsx: true}' --parser typescript-eslint-parser
```

A few questions I have:
- I'm not very knowledgable about other downstream consumers of this parser - will this impact them negatively? In other words, should this only happen when `{ isParseForESLint: true }`, or would this be a welcome change for others in the community as well?
- To make it safer, we could have the passed in config always override the inferred value, however, as I mentioned, I can't think of a situation where this would be the desirable behavior. Doing so would also make existing configurations that have this set continue to do the wrong thing.
- I don't see a way around making this a breaking change, since even if we override it the behavior could be different for someone who does not have `parserOptions.ecmaFeatures.jsx` defined in their configuration. How does the team feel about this?

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -88,15 +88,14 @@ function generateAST(code, options, additionalParsingContext) {
         }
 
         const hasEcmaFeatures = options.ecmaFeatures && typeof options.ecmaFeatures === "object";
-        const hasFilePath = additionalParsingContext.isParseForESLint ? options.filePath !== DEFAULT_ESLINT_FILEPATH : options.filePath;
-        const hasTsxExtension = hasFilePath && /.tsx$/.test(options.filePath);
 
         // Allows user to parse a string of text passed on the command line in JSX mode.
         if (hasEcmaFeatures) {
-            if (typeof options.ecmaFeatures.jsx !== "undefined") {
-                extra.ecmaFeatures.jsx = options.ecmaFeatures.jsx;
-            }
+            extra.ecmaFeatures.jsx = options.ecmaFeatures.jsx;
         }
+
+        const hasFilePath = additionalParsingContext.isParseForESLint ? options.filePath !== DEFAULT_ESLINT_FILEPATH : options.filePath;
+        const hasTsxExtension = hasFilePath && /.tsx$/.test(options.filePath);
 
         // Infer whether or not the parser should parse in "JSX mode" or not.
         // This will override the parserOptions.ecmaFeatures.jsx config option if a filePath is provided.
--- a/parser.js
+++ b/parser.js
@@ -17,8 +17,12 @@ const SUPPORTED_TYPESCRIPT_VERSIONS = require("./package.json").devDependencies.
 const ACTIVE_TYPESCRIPT_VERSION = ts.version;
 const isRunningSupportedTypeScriptVersion = semver.satisfies(ACTIVE_TYPESCRIPT_VERSION, SUPPORTED_TYPESCRIPT_VERSIONS);
 
+const WARNING_BORDER = "=============";
+const DEFAULT_ESLINT_FILEPATH = "<text>";
+
 let extra;
 let warnedAboutTSVersion = false;
+let warnedAboutJSXOverride = false;
 
 /**
  * Resets the extra config object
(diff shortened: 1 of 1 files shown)


shawnrice:
The reasons not to distinguish `ts` from `tsx` are the same reasons found in discussions about a `jsx` file extension. For a lively debate, see the discussion at: https://github.com/airbnb/javascript/pull/985. 

I can understand having a lint rule that requires the `tsx` file extension, but making it part of the parser effectively enables what should be a lint rule, and so it does violate a separation of concerns between the tool and its configuration.

kaicataldo:
@shawnrice It's actually not quite the same thing. In TypeScript, `.tsx` is a requirement for using JSX in your code (unlike when compiling with, say, Babel). See [the official docs](https://www.typescriptlang.org/docs/handbook/jsx.html) for more details.

corbinu:
@kaicataldo

I think that this point breaking changes are pretty expected here as things are still moving pretty fast. I at least am totally ok with this.

JamesHenry:
Very cool, thanks for exploring this @kaicataldo!

@azz I think this will only benefit prettier, right? We could potentially get rid the of the heuristics in https://github.com/prettier/prettier/blob/master/src/language-js/parser-typescript.js

Prettier is the only known programmatic consumer at this point.

azz:
Currently it is possible to use Prettier without passing a file path: 

```
cat foo.tsx | prettier --parser typescript 
```

So we'd have to keep the try tsx then ts behaviour at least in that case, but it would be a welcome improvement.

kaicataldo:
How are we feeling about this?

JamesHenry:
Other than updating it to be a breaking change I think it is good to go!

kaicataldo:
Sorry I lost track of this. Besides for updating the commit message/PR title, what else needs to be done to update it to a breaking change? 

Also, it looks like snapshot tests are failing. My guess is that it has something to do with filename extensions of fixture files, I'll take a look and see what's going on.

parser.js
```diff
@@ -73,17 +77,31 @@ function generateAST(code, options, additionalParsingContext) {
...
+
         if (typeof options.tolerant === "boolean" && options.tolerant) {
             extra.errors = [];
         }
 
-        if (options.ecmaFeatures && typeof options.ecmaFeatures === "object") {
-            // pass through jsx option
-            extra.ecmaFeatures.jsx = options.ecmaFeatures.jsx;
+        const hasEcmaFeatures = options.ecmaFeatures && typeof options.ecmaFeatures === "object";
+        const hasFilePath = options.filePath !== DEFAULT_FILEPATH;
```
kaicataldo: Actually, should this be `options.filePath && (options.filePath !== DEFAULT_FILEPATH)`? If no file path is provided, this would currently still resolve to true.

azz:
If the snapshots fail it's still a breaking change (see my comment above)

kaicataldo:
@azz  Mind clarifying what you mean? I thought we decided this was a breaking change.

JamesHenry:
@kaicataldo Since this was first created I added a lot more sophisticated integration testing: https://github.com/eslint/typescript-eslint-parser#integration-tests

Please could you add some for this? You can ping me in ESLint chat if anything is unclear

JamesHenry:
@kaicataldo If you are interested in reopening this against the new project, please feel free, otherwise I will happily reimplement this myself of there.

Thanks again!

PR #480, opened 2018-06-04
Breaking: Support TypeScript 2.9

JamesHenry:
As per our existing conventions, this change is marked as Breaking despite the lack of breaking code changes. This is to be explicit to users that the change is significant within their setup. Also this branch is protected as usual, so upstream master changes (if any) will need to be merged into this branch.


JamesHenry:
Thanks again, @ikatyang!

PR #510, opened 2018-09-03
Chore: Allow dependencies to dedupe with eslint dependencies

realityking:
Two parts to this PR:

- Allow semver range for dependencies. This matches eslint
- Use the full lodash package (as used by eslint) instead of just the unescape module

### DIFF
--- a/lib/node-utils.js
+++ b/lib/node-utils.js
@@ -12,7 +12,7 @@
 //------------------------------------------------------------------------------
 
 const ts = require("typescript"),
-    unescape = require("lodash.unescape");
+    unescape = require("lodash/unescape");
 
 //------------------------------------------------------------------------------
 // Private
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "betarelease": "eslint-prerelease beta"
   },
   "dependencies": {
-    "lodash.unescape": "4.0.1",
-    "semver": "5.5.0"
+    "lodash.unescape": "^4.0.1",
+    "semver": "^5.5.0"
   },
   "peerDependencies": {
     "typescript": "*"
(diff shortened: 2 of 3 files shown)


JamesHenry:
@realityking It would be useful if you could start with an explanation of your current issue with status quo, and why this solves it for you.

Simply including dependencies because another project does is not a good reason to make a change IMO, but I may be missing something.

Also, please bear in mind that this project is used by more than just ESLint

realityking:
@JamesHenry Started out checking why this module didn't pull in the latest version of semver, deduping with other packages, esp. eslint since it comes from the same "family".

TBH I'm not married to the second commit in this PR but I'd love for the first one to be merged

realityking:
No idea why the commit-message bot fails, as far as I can tell I've followed the docs: https://eslint.org/docs/developer-guide/contributing/pull-requests#step-2-make-your-changes

j-f1:
@realityking I think the PR title has to follow the message rules if you have more than one commit.

realityking:
@j-f1 That was it. Thank you!

JamesHenry:
@realityking I am grateful for you contributing, but I'm afraid I still haven't understood the reason to make this change. As far as I have understood from you above, there is not any behaviour of this library that this PR fixes/improves, it just streamlines your personal node_modules directory based on your usage.

Having ranges in `dependencies` opens you up to having a "range" of behaviours, depending on how and when the library is installed, so unless there is a compelling reason to make the change, I think we should stick with the current approach. Yes, strict adherence to semver should minimise the threat of breaking changes occurring in the ranges in this PR, but that relies on package authors not making mistakes.

If that summary is incorrect please feel free to reopen and disucuss further. Thanks again!

PR #515, opened 2018-09-22
Breaking: Use typescript-estree for parsing

JamesHenry:
_Note: There are no known user-facing breaking changes introduced by this PR, but I think it is worth marking it as such to make it clear that a major refactoring has taken place_

As approved by the ESLint TSC, this PR:

- Replaces the parsing-related internals with the new `typescript-estree` package - a drop-in replacement
- Removes the now irrelevant babylon ast-alignment-tests
- Removes any unnecessary `dependencies` or `devDependencies`
- Updates the README

As mentioned on gitter, there was only ever one case where we customized the AST for ESLint's benefit previously, and so I have carried over that behaviour by leveraging the ESLint traverser in a post-processing step. Please let me know if there is a better way to maintain this behaviour (without needing to go as far as adding a custom option in `typescript-estree`).

Thank you ❤️


JamesHenry:
Btw I would be really happy for other ESLint folks to be invited into the new project!

@kaicataldo You have helped out on this project in various ways, I will add you initially, but feel free to reach out if you would rather not be involved directly

parser.js
```diff
@@ -8,189 +8,29 @@
...
 exports.version = require("./package.json").version;
 
-exports.parse = function parse(code, options) {
-    return generateAST(code, options, { isParseForESLint: false });
-};
-
 exports.parseForESLint = function parseForESLint(code, options) {
-    const ast = generateAST(code, options, { isParseForESLint: true });
+    const ast = parse(code, options);
+    traverser.traverse(ast, {
```
kaicataldo: Can you refresh my memory again about when a function can exist without a body - is that only for type definitions?

Either way, if this shouldn't belong in the parser, I think this is a fine solution for now :+1:

parser.js
```diff
@@ -8,189 +8,29 @@
...
-    }, compilerHost);
-
-    const ast = program.getSourceFile(FILENAME);
-
-    extra.code = code;
-    return convert(ast, extra);
-}
+const parse = require("typescript-estree").parse;
+const astNodeTypes = require("typescript-estree").AST_NODE_TYPES;
+/* eslint-disable */
```
kaicataldo: What is this `eslint-disable` comment needed for?

JamesHenry: From memory I think cause I am deep importing from a node_module

parser.js
JamesHenry: Here are some examples:

```ts
declare function f1(x: number): void;

interface Foo {
    bar(): boolean;
}
```

JamesHenry:
@kaicataldo 

> Thanks for doing this work! I have two questions, but I trust your judgment on this one and don't  think they're blockers.

> Regarding adding me to the repo: if you're looking for someone to commit to maintaining the project right now, I'm unsure of what time commitments I can make at the moment. That being said, I'm happy to be involved and help out where I can!

No, zero expectations for contributions! I just wanted to make it clear that ESLint can be represented :)

parser.js
JamesHenry: @platinumazure This was a great thing to flag because the underlying issue was that with this change ESLint went from being just a devDep to a full dep of the project, there is no need for the disable comment at all now that I have moved it into deps

JamesHenry:
@platinumazure No it never actually had any parserServices, they were planned for a while, I even had a PR open, but they never went in, so no changes introduced by this PR.

Feedback addressed, PTAL!

JamesHenry:
Just FYI I will go through and do some comprehensive issue triage once this change happens, migrating any purely parsing related issues to the other repo

JamesHenry:
Right now all the TypeScript dependencies are based on peers, so it would be a single version based on what the user installed, but it is a good point to raise in case that ever changed.

I would definitely be happy to expose the ts.Program instance from the `typescript-estree` and then that (or a subset of it) could be further exposed via parserServices here. Just let me know (and I have already added @kaicataldo as a collaborator, happy to add anyone else from the team)

Thanks to all of you for the reviews!

PR #530, opened 2018-10-25
Docs: add options to README

trusktr:
Are these options inherited from somewhere? If so, then maybe we should just mention that, and only list ones that are specific to this parser.


trusktr:
I signed the CLA. I also updated the README. How does that look?

kaicataldo:
Thanks for contributing!

Issue #503, opened 2018-07-25
JSX fragment causes false positives with no-multi-str


nickmccurdy:
**What version of TypeScript are you using?**
2.9.2

**What version of `typescript-eslint-parser` are you using?**
16.0.1

**What code were you trying to parse?**

```ts
export default function Home() {
  return (
    <>
      {description}
      <br />
      <Options labels={getOptions(questions)} />
    </>
  )
}
```

**What did you expect to happen?**
No errors, because `<></>` is short for `<React.Fragment></React.Fragment>` and `React` is already imported.

**What happened?**
`Multiline support is limited to browsers supporting ES5 only  no-multi-str`

nickmccurdy:
I couldn't figure out how to get it to behave with the shorthand, so I disabled the rule. With the number of exceptions to rules like this in this parser, I think it would be really useful to include a shared config that disables rules conflicting with TypeScript and the parser implementation.

ThaJay:
subscribing, I als have both issues (quotes and no-multi-str get triggered)

With just jsx the quotes rule got triggered and I moved to babel/quotes to fix this.
Now with tsx both rules get triggered (and thus the babel/quotes rule also does not work for ts)

lostfictions:
Looks like this might be fixed (via #511, now brought in as https://github.com/JamesHenry/typescript-estree/pull/15)

infctr:
Is the dependency going to be updated? It currently has `"typescript-estree": "2.1.0"` while `4.0.0` is the latest version

alangpierce:
https://github.com/JamesHenry/typescript-estree/pull/15 was published in typescript-estree 2.1.0, so I was able to get fragments working by updating typescript-eslint-parser to latest. (Looks like the 4.0.0 is just from two breaking change commits since the fragments commit, not sure how important those are to pull in.)

PR #538, opened 2018-11-01
Breaking: typescript-estree v5

kaicataldo:
Following up on https://github.com/JamesHenry/typescript-estree/pull/22, this updates to the latest version of typescript-estree.

Looks like it might also fix https://github.com/eslint/typescript-eslint-parser/issues/503.

I think we should remove support for `ecmaFeatures` in the options and add individual feature flags if we need them in the future (which seems unlikely). Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,9 @@ npm install --save-dev typescript-eslint-parser
 In your ESLint configuration file, set the `parser` property:
 
 ```json
-"parser": "typescript-eslint-parser"
+{
+    "parser": "typescript-eslint-parser"
+}
 ```
 
 There is sometimes an incorrect assumption that the parser itself is what does everything necessary to facilitate the use of ESLint with TypeScript. In actuality, it is the combination of the parser _and_ one or more plugins which allow you to maximize your usage of ESLint with TypeScript.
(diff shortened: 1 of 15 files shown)


tools/test-utils.js
```diff
@@ -41,7 +41,16 @@ function getRaw(ast) {
...
  * @returns {Function} callback for Jest test() block
  */
-function createSnapshotTestBlock(code, config) {
+function createSnapshotTestBlock(code, config = {}) {
+    const defaultConfig = {
+        loc: true,
+        range: true,
+        raw: true,
+        tokens: true,
+        comment: true,
```
kaicataldo: Not all the tests had `comment: true`, which is what is causing a lot of diffs in the snapshots. I personally think it's fine to have this always set.

README.md
```diff
@@ -30,15 +32,23 @@ Instead, you also need to make use of one more plugins which will add or extend
 
 By far the most common case will be installing the [eslint-plugin-typescript](https://github.com/nzakas/eslint-plugin-typescript) plugin, but there are also other relevant options available such a [eslint-plugin-tslint](https://github.com/JamesHenry/eslint-plugin-tslint).
 
-## Options
+## Configuration
+
+The following additional configuration options are available by specifying them in [`parserOptions`](https://eslint.org/docs/user-guide/configuring#specifying-parser-options) in your ESLint configuration file.
+
+**`jsx`** - default `false`. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
```
kaicataldo: There are technically other options (`range`, `loc`, `comment`, etc.), but these aren't particularly useful for the end user, since ESLint supplies them. I think we should document the API (`parse()`, `parseForESLint()`, etc.) and document those options there.

platinumazure:
FYI: In another issue, I promised I would cut a major release (to change ESLint from a dependency to a peerDependency/devDependency) at some point. I think it would be great to include both changes in the next major release. I'm okay with handling the release, but if you would prefer to do it, let me know if you have questions.

So in summary: Please merge this whenever you feel ready, but please consult me or at least check the package.json file before cutting a release. Thanks!

kaicataldo:
@platinumazure I'm going to merge this now. Happy to take on those other changes and cut a release!

kaicataldo:
PR [here](https://github.com/eslint/typescript-eslint-parser/pull/541)

Issue #488, opened 2018-06-18
no-unused-vars should understand imported type usage


Snugug:
**What version of TypeScript are you using?**
2.9.1

**What version of `typescript-eslint-parser` are you using?**
16.0.0

**What code were you trying to parse?**

```ts
import { Application } from 'express';

app().then((server:Application) => {

});
```

**What did you expect to happen?**
With rule `no-unused-vars`, I expect `Application` to come back as being used as it's being imported and used as a type

**What happened?**
`  1:10  error  'Application' is defined but never used  no-unused-vars`

st-sloth:
This issue duplicates #457 that is closed because this problem is not a parsing but a rule issue.

armano2:
To make it work i suggest using https://github.com/bradzacher/eslint-plugin-typescript/blob/master/docs/rules/no-unused-vars.md

JamesHenry:
Using the plugin probably makes the most sense here, ESLint will not be able to understand many TS specific concerns natively.

Please feel free to open any other issues on the new project https://github.com/typescript-eslint/typescript-eslint

Thanks!

PR #540, opened 2018-11-05
Update: add proper scope analysis (fixes #535)

### LINKED ISSUE #416: False positive on `no-undef` with mixed JS & TS codebase

2.6.1

11.0.0

```
export type SomeThing = {
    id: string;
}
```

No lint error

**What happened?**

False positive on `no-undef` eslint core rule.

It basically means we can not use type definitions without disabling that rule (globablly or locally)

This has already been discussed in https://github.com/eslint/typescript-eslint-parser/issues/77

It really becomes tricky when the project mixes JS and TS code.

But we haven't found any viable solution yet, apart applying different configuration to the various file types.

It would be nice if the eslint core rules would be overridable by the Typescript parser usage somehow I guess.

*BUT this is really only an issue for mixed JS/TS codebases, because*: 

> let's keep in mind this is easily fixable if your code is Typescript only, 
> since you can boldly disable this core rule as the TS compiler should complain anyway about a 
> missing definition of something :)
> It really only becomes an issue when you want to take care of JS code using the same lint config.

### LINKED ISSUE #435: Interface keys can falsely violate `no-use-before-define`
2.6.2

11.0.0

```ts
interface Foo {
    bar: string
}

const bar = 'blah'
```

The code to be considered valid from the perspective of `no-use-before-define`, or the README to call out that this is an unsupported case.

**What happened?**
The key `bar` in the interface on line 2 was flagged as being used before defined. If I removed the later `const bar...` assignment, the lint error went away.

### LINKED ISSUE #437: Interface produces no-undef errors
2.6.2

12.0.0

```ts
interface Runnable {
  run (): Result
  toString (): string
}
```

Pass with no errors.

**What happened?**
```
> eslint **/*.{j,t}s


/Users/nick/Repos/purespec/matchers/Runnable.ts
  1:11  error  'Runnable' is not defined  no-undef
  2:3   error  'run' is not defined       no-undef
  2:11  error  'Result' is not defined    no-undef

✖ 3 problems (3 errors, 0 warnings)
```

### LINKED ISSUE #443: no-redeclare: false positives for type definitions
2.6.1

12.0.0

```ts
const Foo = 1;
// unexected error: 'Foo' is already defined. (no-redeclare)
type Foo = 1;
```

We've decided to disable this ESLint rule as TypeScript also provides similar functionality anyway, but logging this just for others/reference.

For context, this pattern is common when using [unionize](https://github.com/pelotom/unionize):

``` ts
import { unionize, ofType } from 'unionize'

export const Action = unionize({
  ADD_TODO:                ofType<{ id: string; text: string }>(),
  SET_VISIBILITY_FILTER:   ofType<'SHOW_ALL' | 'SHOW_ACTIVE' | 'SHOW_COMPLETED'>(),
  TOGGLE_TODO:             ofType<{ id: string }>(),
});

export type Action = typeof Action._Union;
```

### LINKED ISSUE #459: no-shadow if a type alias and function argument share the same name
2.7.1

14.0.0

```ts
type foo = any;
function bar(foo: any) {}
```

Not a `no-shadow` error.

**What happened?**

```
2:14  error  'foo' is already declared in the upper scope  no-shadow
```

In contrast, this works as expected (no such error):

```ts
type foo = any;
let foo: any;
```

### LINKED ISSUE #466: What's the recommended way to work around "is not defined" for type definitions?
Hello, how can I work around the error

```
  ✘  http://eslint.org/docs/rules/no-undef                'NodeListOf' is not defined
  src/util.ts:10:60
      const links = document.querySelectorAll( selector ) as NodeListOf<HTMLElement>
```

? Do I have to use a comment to tell eslint to ignore that error on that line? Is there any other better workaround?

### LINKED ISSUE #471: Class fields cause false positive no-undef
`2.7.2`, `2.8.1`

`15.0.0`

```ts
class X {
  field = {} // [eslint] 'field' is not defined. (no-undef)
}
```

I did not expect any error

**What happened?**
I get `[eslint] 'field' is not defined. (no-undef)`

### LINKED ISSUE #476: TSTypeReference does not emit for React.SFC

2.8.1

0.12.0

```ts
export interface PriceProps {
  // Some props
}

export const Price: React.SFC<PriceProps> = function Price(props) {}
```

I'm trying to write a rule that verifies the generics in `React.SFC` are correct, however, the `TSTypeReference` for `React.SFC` is never emitted, so I'm unable to capture and and write a rule around it.

For example, this doesn't even log.

```
TSTypeReference(node) {
  console.log(node);
},
```

**What happened?**

Doesn't work.

### LINKED ISSUE #487: no-restricted-globals: false positive on class property
2.9.1

16.0.0

With `no-restricted-globals` rule enabled (`"no-restricted-globals": [2, "status"]`):
```ts
export default class Test {
	private status: string;

	getStatus() {
		return this.status;
	}
}
```

Lint completes without error.

**What happened?**
False positive for `no-restricted-globals`:

    2:10  error  Unexpected use of 'status'  no-restricted-globals

### LINKED ISSUE #535: no-dupe-args and no-redeclare false positives in 20.1.0
3.1.4

20.1.0

```ts
function foo({ bar }: { bar: string }) {
	console.log(bar);
}
```

No errors.

**What happened?**
```
  1:1   error  Duplicate param 'bar'         no-dupe-args
  1:25  error  'bar' is already defined      no-redeclare
```

### LINKED ISSUE #536: False positive for unused vars used as decorators of properties
3.1.4

20.1.0

```ts
import {
  action,
  observable,
  useStrict,
  runInAction,
  IObservableArray,
} from 'mobx';
import orderReadService from '@client/model/purchase/orderReadService';
import {
  BizPurchaseOrderTO,
  PageTO,
  BizQueryOrderConditionTO,
} from '@client/types/thrift/purchase';
import { PURCHASE_ORDER_TYPE_ENUM } from '@constants/order';

useStrict(true);
export default class ListModalStore {
  // order list
  @observable
  orderList: IObservableArray<BizPurchaseOrderTO> = observable([]);
  @observable
  orderListPageInfo: PageTO = { currentPage: 1, perPage: 10, total: '5' };
}
```

no eslint error for the 'observable' which is imported from mobx.

**What happened?**

[eslint] 'observable' is defined but never used.

This is happened for the 'observable' which is imported from mobx.

### LINKED ISSUE #545: Imported interfaces destructured in function argument gives no-shadow
Error occurred after upgrade from `typescript-eslint-parser^20.0.0`.
Potentially related to #540.

3.1.6

20.1.1

```ts
import { Foo } from "./Foo";

function foo(args: Foo) {} // No error
function bar([...args]: Foo) {} // No error
function baz({ ...args }: Foo) {} //  'Foo' is already declared in the upper scope

const foo = Foo; // No error
const [...foo] = Foo; // No error
const { ...foo } = Foo; // No error
```

mysticatea:
This PR adds proper scope analysis by returning [`scopeManager`](https://eslint.org/docs/developer-guide/working-with-custom-parsers) object from `parseForESLint` function.

The `scopeManager` doesn't contain any types. I think that we can provide `parserServices` to get types, but I don't contain it in this PR.

<details>
<summary>This fixes #416, #435, #437, #443, #459, #466, #471, #476, #487, #535, #536, and #545.</summary>

- Fixes #435
- Fixes #437
- Fixes #443
- Fixes #459
- Fixes #466
- Fixes #471
- Fixes #476
- Fixes #487
- Fixes #535
- Fixes #536
- Fixes #545

</details>


mysticatea:
I'm a newbie about jest. Please advise me on how to write tests with jest.
Currently, I use snapshot tests with the scope tree after I convert circular references to `{ $id: X ... }` and `{ $ref: X }`.

tests/fixtures/scope-analysis/decorators.ts
```diff
@@ -0,0 +1,13 @@
+function dec(target: any) {
+}
+function gec() {
+    return (target: any, proeprtyKey: string) => {}
```
platinumazure: Typo: "proeprtyKey"

tests/fixtures/scope-analysis/typeof-in-type-parameters.ts
```diff
@@ -0,0 +1,3 @@
+function g<T extends typeof g>(g: T): number {
```
platinumazure: In `<T extends typeof g>`, is `g` the function name or the parameter name? Could one of the `g` identifiers be renamed for clarity, or are these supposed to be identical symbols in different symbol tables?

tests/fixtures/scope-analysis/typeof-in-var.ts
```diff
@@ -0,0 +1,3 @@
+var obj = { value: 1 }
+var obj2: typeof obj = { value: 2 }
+var { value }: typeof obj = { value: 2 }
```
platinumazure: Is there an array destructuring case we could add here?

tests/fixtures/scope-analysis/typeof-in-type-parameters.ts
mysticatea: The function name.

This test checks the parameter name doesn't shadow the function name for type parameters, as well.

mysticatea:
I'm wondering to distinguish `declare global {}` blocks :thinking: (https://github.com/JamesHenry/typescript-estree/issues/27)

mysticatea:
I'm wondering on #488.

`ImportDeclaration`s can import type stuff, but we cannot distinguish whether the imported item is a variable or a type. So I cannot ignore types on `ImportDeclaration` node :thinking:

mysticatea:
I have done except about #488.

mysticatea:
I hope to separate PR for #488.

JamesHenry:
I'll take a look at this tonight, thanks a lot @mysticatea!

tests/lib/scope-analysis.js
```diff
@@ -0,0 +1,355 @@
...
+`;
+        const config = {
+            parser: "typescript-eslint-parser",
+            rules: {
+                "no-unused-vars": "error"
+            }
+        };
+        const messages = linter.verify(code, config, { filename: "issue.ts" });
+
+        assert.deepStrictEqual(messages, []);
```
JamesHenry: How come you are frequently using `assert` rather than jest's `expect()` within the jest `test()` blocks?

I can't say I have ever seen or done this so just calling it out in case there is any chance it is masking some issues/causing false positives...

JamesHenry:
Amazing work, @mysticatea! I just had the one question about the tests. I think your approach to the serialization is reasonable, although you can also create a custom matcher for jest as an alternative: https://jestjs.io/docs/en/expect#custom-matchers-api

JamesHenry:
I don't know if you noticed the pattern I established for running integration tests within docker containers in this project.

With so many potentially closed issues here, it might be nice to add them in as one or more integration tests? Let me know what you think and if you need any help with that

tests/lib/scope-analysis.js
mysticatea: Oops. I have used `mocha` always, so I couldn't have the idea that `jest` has assertion functions. I will fix it.

mysticatea:
:thinking: I need time a bit to understand how those integration tests work since I'm not familiar with docker (and I'm on Windows 7).

JamesHenry:
Ok no problem regarding the integration tests, I can always add them in a follow up

mysticatea:
I removed `assert()`.

lvivski:
Why is it still blocked? It would be great to have this PR merged. Right now we have a add a ton of exceptions in config because of all the false-positives

platinumazure:
Hi @ lvivski, there are a couple of people who still need to review this, including me.

This is open source and we all work as volunteers. We sometimes have other obligations in our lives. I'm sorry that this sometimes results in inconvenience for users who are eagerly waiting for a feature to get merged.

lvivski:
@platinumazure sorry, I didn't mean to blame anyone but rather wanted to understand if there's anything that I can help with.

platinumazure:
@lvivski Thank you for your understanding. I apologize for misinterpreting your previous comment.

I've just taken a look and I only need to double-check one thing before I can leave my review-- hopefully I'll be able to approve after that check and then we can get this in soon!

lvivski:
Thank you so much for taking a look into it!

mysticatea:
@platinumazure Thank you very much!

Issue #534, opened 2018-10-31
20.1.0 release contains a breaking change


alexzherdev:
**What version of TypeScript are you using?**
3.1.3

**What version of `typescript-eslint-parser` are you using?**
20.1.0

**What did you expect to happen?**
When upgrading from 20.0.0 to 20.1.0 there are no breaking changes.

**What happened?**
As discussed in https://github.com/yannickcr/eslint-plugin-react/pull/2032, changing a dep to peerDep in #526 is a breaking change.

platinumazure:
My apologies, I mislabeled the commit as a chore rather than breaking.

I'll revert the commit, cut a patch release, then redo the commit as breaking and cut a major release. I'm traveling today, so I greatly appreciate your patience as I find time to do this in between airline security, etc. Thanks for your understanding!

platinumazure:
I think I've published typescript-eslint-parser@20.1.1 reverting the peer dependency change. (I say "I think" because this is the first time I've used our new release pipeline with npm 2FA. I do see commits on GitHub and an artifact on npm, so hopefully all went well.)

I'm about to fly now, so I'll do the second part (apply the commit as breaking and cut major release) later today.

PR #541, opened 2018-11-07
Breaking: make eslint a peerDep/devDep

kaicataldo:
refs https://github.com/eslint/typescript-eslint-parser/pull/538#issuecomment-436252936
refs https://github.com/eslint/typescript-eslint-parser/issues/534

This makes `eslint` a `peerDependency` and will be batched with https://github.com/eslint/typescript-eslint-parser/pull/538 in the next major release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,16 +17,6 @@
     "url": "https://github.com/eslint/typescript-eslint-parser/issues"
   },
   "license": "BSD-2-Clause",
-  "devDependencies": {
-    "eslint-config-eslint": "4.0.0",
-    "eslint-plugin-node": "6.0.1",
-    "eslint-release": "1.1.0",
-    "jest": "23.1.0",
-    "npm-license": "0.3.3",
-    "shelljs": "0.8.2",
-    "shelljs-nodecli": "0.1.1",
-    "typescript": "~3.1.1"
-  },
   "keywords": [
     "ast",
     "ecmascript",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -48,13 +38,25 @@
     "generate-rcrelease": "eslint-generate-prerelease rc",
     "publish-release": "eslint-publish-release"
   },
+  "peerDependencies": {
+    "typescript": "*",
+    "eslint": ">=4.19.1",
+    "eslint-visitor-keys": ">=1.0.0"
```
kaicataldo: Any reason why we shouldn't also rely on the version of `eslint-visitor-keys` that's already installed as a dep of `eslint`?

platinumazure: Not sure I follow. Are you asking why we don't pull in eslint-visitor-keys as a dependency, or are you saying the version here is inaccurate? (Or something else entirely?)

kaicataldo: Sorry, let me try to clarify: `eslint-visitor-keys` is currently a dependency. However, `eslint` already includes it as a direct dependency, leading to the possibility of there being two different versions of `eslint-visitor-keys` being used (one in `eslint` and one in `typescript-eslint-parser`). This could be mitigated by requiring `eslint-visitor-keys` as a `peerDependency` instead. I just wasn't sure if there was a reason we might actually want that behavior.

not-an-aardvark: I think it would be better to leave it as a `dependency`. Deduplicating dependencies seems like the package manager's job, not the package authors' job. To my knowledge, yarn and recent versions of npm would only install two different versions of a package if there's an incompatibility between the specified versions.

Additionally, if we made `eslint-visitor-keys` a `peerDependency`, then users would need to manually install it. (If they didn't manually install it, then their builds might unexpectedly break if `eslint` decided to change its implementation and not depend on `eslint-visitor-keys`.) So duplicate packages might end up appearing anyway if the package manager didn't deduplicate them. It seems like by using a `peerDependency` here we would be adding installation effort without much benefit.

Changing the range to `>=1.0.0` would also cause this package to break if we made breaking changes to `eslint-visitor-keys`, which seems non-ideal.

I think the only scenario where it would make sense to use a `peerDependency` here would be if we wanted the behavior of `typescript-eslint-parser` to be customizable by having the user install different versions of `eslint-visitor-keys` for it to use. But that doesn't seem to be the case here.

kaicataldo: That's fair! You've convinced me. I'll revert this change.

kaicataldo:
@platinumazure @not-an-aardvark Thanks for looking this over! Ready for re-review.

PR #543, opened 2018-11-08
Breaking: switch 'jsx' option by filename (fixes #517)

### LINKED ISSUE #517: Known issue: Ambiguous syntax in JSX context
In this parser we have to use the `parserOptions.jsx` flag to determine whether or not to configure the actual TypeScript compiler to parse the source as JSX.

When the TypeScript compiler parses certain syntax in "JSX mode" it is ambiguous and could be a JSXElement as far as it's concerned.

This is the same reason you cannot use the `<>` syntax for type assertions in TSX files, you have to use the `as` syntax.

Essentially then, if you want to have `jsx: true`, you have to accept that you are always in "JSX mode" in TypeScript, working in the same way as if all of your source files were `.tsx`, and so you cannot use any ambiguous constructs.

Right now the DX is not at all clear when you have configured `jsx: true` in your ESLint config and then use ambiguous TypeScript constructs. You will just end up with a random parse error from the TypeScript compiler.

Ideally, we would be able to provide clearer error messaging, or infer the jsx setting itself from the file extension (.ts vs .tsx)

You can usually reproduce the behaviour using the TypeScript compiler directly on: https://astexplorer.net/

mysticatea:
This PR makes the parser setting `jsx` option by filename automatically.

- `*.ts` ... it sets `jsx: false`.
- `*.tsx` ... it sets `jsx: true`.
- Otherwise (E.g., `*.js`, not provided, etc...) ... respects the given `jsx` option.

(Edited at `2018-11-08 20:52Z`)

Fixes #517.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,12 @@ By far the most common case will be installing the [eslint-plugin-typescript](ht
 
 The following additional configuration options are available by specifying them in [`parserOptions`](https://eslint.org/docs/user-guide/configuring#specifying-parser-options) in your ESLint configuration file.
 
-**`jsx`** - default `false`. It's `true` on `*.tsx` files automatically. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
+- **`jsx`** - default `false`. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
+    - It's `false` on `*.ts` files regardless of this option.
+    - It's `true` on `*.tsx` files regardless of this option.
+    - Otherwise, it respects this option.
 
-**`useJSXTextNode`** - default `false`. The JSX AST changed the node type for string literals inside a JSX Element from `Literal` to `JSXText`. When value is `true`, these nodes will be parsed as type `JSXText`. When value is `false`, these nodes will be parsed as type `Literal`.
+- **`useJSXTextNode`** - default `false`. The JSX AST changed the node type for string literals inside a JSX Element from `Literal` to `JSXText`. When value is `true`, these nodes will be parsed as type `JSXText`. When value is `false`, these nodes will be parsed as type `Literal`.
 
 ### .eslintrc.json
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -36,7 +36,7 @@ By far the most common case will be installing the [eslint-plugin-typescript](ht
 
 The following additional configuration options are available by specifying them in [`parserOptions`](https://eslint.org/docs/user-guide/configuring#specifying-parser-options) in your ESLint configuration file.
 
-**`jsx`** - default `false`. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
+**`jsx`** - default `false`. It's `true` on `*.tsx` files automatically. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
```
kaicataldo: I wonder if we can make this clearer by explaining that the parser will auto detect this and this config option is only for when a file name is not provided.

parser.js
```diff
@@ -20,6 +20,10 @@ const visitorKeys = require("./visitor-keys");
 exports.version = require("./package.json").version;
 
 exports.parseForESLint = function parseForESLint(code, options) {
+    if (options && typeof options.filePath === "string" && options.filePath.endsWith(".tsx")) {
```
kaicataldo: I wonder if the behavior should be that it only auto-detects when the `jsx` option is not defined? I think it could be confusing for a user to have `jsx: false` and still have it parse JSX.

mysticatea: Hm. After I re-thought it:

On TypeScript, we cannot enable jsx on `*.ts` files and cannot disable jsx on `*.tsx` files.
So probably makes sense if it uses the `jsx` option only when filename is not provided.

mysticatea: Ah... No, when neither `*.ts`  or `*.tsx` then it use the given `jsx` option.

README.md
mysticatea: I updated it.

mysticatea:
@JamesHenry Thank you for the review.

The intention of this PR's spec is:

- It's always `"auto"` essentially because we cannot configure it on `tsc` -- we cannot use JSX syntax on `*.ts` files and cannot disable JSX syntax on `*.tsx` files on `tsc`. So the `jsx` option on TypeScript files doesn't make sense.
- But people may use this parser on other files such as `*.js`. Or if people use `stdin`, the filename is unknown. This `jsx` option is provided for such a case.  
  E.g., `curl a_url | eslint --stdin --perser-options jsx:true`

kaicataldo:
Thanks for contributing!

PR #544, opened 2018-11-08
Breaking: make 'useJSXTextNode:true' by default

mysticatea:
This PR makes the parser `useJSXTextNode: true` by default because the current ESLint assumes so.

And add a note to README.md: "Please set `false` if you use this parser on ESLint v4."

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The following additional configuration options are available by specifying them 
 
 **`jsx`** - default `false`. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
 
-**`useJSXTextNode`** - default `true`. The JSX AST changed the node type for string literals inside a JSX Element from `Literal` to `JSXText`. When value is `true`, these nodes will be parsed as type `JSXText`. When value is `false`, these nodes will be parsed as type `Literal`. Please set `false` if you use this parser on ESLint v4.
+**`useJSXTextNode`** - default `true`. Please set `false` if you use this parser on ESLint v4. If this is `false`, the parser creates the AST of JSX texts as the legacy style.
 
 ### .eslintrc.json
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ The following additional configuration options are available by specifying them
 
 **`jsx`** - default `false`. Enable parsing JSX when `true`. More details can be found [here](https://www.typescriptlang.org/docs/handbook/jsx.html).
 
-**`useJSXTextNode`** - default `false`. The JSX AST changed the node type for string literals inside a JSX Element from `Literal` to `JSXText`. When value is `true`, these nodes will be parsed as type `JSXText`. When value is `false`, these nodes will be parsed as type `Literal`.
+**`useJSXTextNode`** - default `true`. The JSX AST changed the node type for string literals inside a JSX Element from `Literal` to `JSXText`. When value is `true`, these nodes will be parsed as type `JSXText`. When value is `false`, these nodes will be parsed as type `Literal`. Please set `false` if you use this parser on ESLint v4.
```
platinumazure: I'm wondering if the note at the end of this comment should be made more prominent in some way (e.g., apply formatting, or put it on its own line, or...)

Maybe someone else has a thought on if this should be done?

mysticatea: How about...?

```suggestion
**`useJSXTextNode`** - default `true`. Please set `false` if you use this parser on ESLint v4. If this is `false`, the parser creates the AST of JSX texts as the legacy style.
```

platinumazure: That could work, thanks!

mysticatea: Thank you.

I used the "suggested change β" feature for the first time :)

kaicataldo:
Looks like merging the other PR has caused some merge conflicts, but otherwise this looks ready to go.

kaicataldo:
@mysticatea Actually, should we wait for https://github.com/eslint/typescript-eslint-parser/pull/540 to land before doing a release?

mysticatea:
@kaicataldo I have merged that PR.

PR #547, opened 2018-11-08
Breaking: throw syntax error on incomplete variable declarations (fixes #531)

### LINKED ISSUE #531: got "TypeError: Cannot read property 'loc' of undefined" after entering 'const' in typescript file in vscode
v2.9.1

v16.0.0

```ts
// Put your code here
const
```

no error

**What happened?**
After entering const
```
[Trace - 3:10:47 PM] Received notification 'window/logMessage'.
[Error - 3:10:47 PM] TypeError: Cannot read property 'loc' of undefined
at Object.VariableDeclaration [as listener] (/openSource/koa-typescript-example/node_modules/eslint/lib/rules/indent.js:1317:69)
at Program:exit.listenerCallQueue.filter.forEach.nodeInfo (/openSource/koa-typescript-example/node_modules/eslint/lib/rules/indent.js:1496:55)
at Array.forEach ()
at Program:exit (/openSource/koa-typescript-example/node_modules/eslint/lib/rules/indent.js:1496:26)
at listeners.(anonymous function).forEach.listener (/openSource/koa-typescript-example/node_modules/eslint/lib/util/safe-emitter.js:47:58)
at Array.forEach ()
at Object.emit (/openSource/koa-typescript-example/node_modules/eslint/lib/util/safe-emitter.js:47:38)
at NodeEventGenerator.applySelector (/openSource/koa-typescript-example/node_modules/eslint/lib/util/node-event-generator.js:251:26)
at NodeEventGenerator.applySelectors (/openSource/koa-typescript-example/node_modules/eslint/lib/util/node-event-generator.js:280:22)
at NodeEventGenerator.leaveNode (/openSource/koa-typescript-example/node_modules/eslint/lib/util/node-event-generator.js:303:14)
```

[eslint issue #11023](https://github.com/eslint/eslint/issues/11023)

mysticatea:
This PR makes the parser throwing on incomplete variable declarations.

It's a syntax error on both JavaScript and TypeScript. But TypeScript parser looks so tolerant and triggers that core rules are confusing.


package.json
```diff
@@ -63,6 +63,15 @@
     "collectCoverage": true,
     "coverageReporters": [
       "text-summary"
+    ],
+    "collectCoverageFrom": [
```
mysticatea: I added this configuration of `jest` because `jest` threw a syntax error on new `tests/fixtures/basics/solo-const.src.js` file.

mysticatea:
I'm thinking about it, too.

I'm not sure if this is a spec violation or not because: ESTree defines types (E.g., `VariableDeclaration#declarations` is `Array<VariableDeclarator>`) but ESTree doesn't define restrictions (E.g., `VariableDeclaration#declarations.length` should not be `0`).

Anyway, I should open an issue on typescript-estree repo.

JamesHenry:
Yep we can get that in as an option in the parser 👍

JamesHenry:
See my comment here after Andy from the TypeScript team responded: https://github.com/eslint/typescript-eslint-parser/issues/531#issuecomment-437238535

It should hopefully be possible for us to leverage TypeScript here, as long as we can clarify the behaviour is correct

JamesHenry:
Response from Andy:

> You could open an issue, though this is hard to change because grammar and semantic checking are coupled currently.

Maybe a good next step would be if I add an option to the parser (which will be `false` by default so as to be non-breaking) which checks for results from both:

```ts
program.getSyntacticDiagnostics(file);
program.getSemanticDiagnostics(file);
```

E.g. the option could be 

```
"errorOnTypeScriptSyntaticAndSemanticIssues": false
```

...and we could see if that is something this ESLint parser would to enable?

mysticatea:
I'm sorry for the delay. I have overlooked this.

`errorOnTypeScriptSyntaticAndSemanticIssues` option sounds good to me.

armano2:
`errorOnTypeScriptSyntaticAndSemanticIssues` seems nice
for now i implemented this as rule in https://github.com/bradzacher/eslint-plugin-typescript/pull/245

j-f1:
`throwSemanticErrors`?

JamesHenry:
FYI I added the `errorOnTypeScriptSyntaticAndSemanticIssues` option to `typescript-estree` in version `8.1.0`, so you can opt in to that and it will error in the exact test case you have in this PR

JamesHenry:
Once we merge the latest updates of typescript-estree (https://github.com/eslint/typescript-eslint-parser/pull/596), we can explore surfacing an option for the diagnostics reporting

JamesHenry:
Closing this for now, we'll add in full support for `errorOnTypeScriptSyntaticAndSemanticIssues`

Issue #553, opened 2018-11-17
Crash on running ESLint


ficristo:
**What version of TypeScript are you using?**
3.1.6

**What version of `typescript-eslint-parser` are you using?**
21.0.1

**What code were you trying to parse?**
See https://github.com/quadre-code/quadre/pull/103
I cannot hunt down to specific code...

**What did you expect to happen?**
Eslint should pass.

**What happened?**
```
TypeError: Cannot read property 'type' of undefined
    at isForInRef (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:406:24)
    at variable.references.some.ref (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:443:21)
    at Array.some (<anonymous>)
    at isUsedVariable (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:442:40)
    at collectUnusedVariables (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:565:26)
    at collectUnusedVariables (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:572:17)
    at collectUnusedVariables (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:572:17)
    at Program:exit (F:\github\quadre\node_modules\eslint\lib\rules\no-unused-vars.js:617:36)
    at listeners.(anonymous function).forEach.listener (F:\github\quadre\node_modules\eslint\lib\util\safe-emitter.js:47:58)
    at Array.forEach (<anonymous>)
```

**Some more info**
At the moment I cannot update ESLint to version 5.*
To see the error on my project I run `grunt eslint --stack --debug`.
Is it possible the crash is because I do not use eslint-plugin-typescript and its typescript/no-unused-vars rule? Nonetheless I would expect an ESLint error and not a crash.

mysticatea:
Thank you for this issue.

However, we cannot guarantee that ESLint core rules work properly on TypeScript syntax because the rules don't know TypeScript syntax. It's the reason that `eslint-plugin-typescript` has some copies of core rules.

I'd like to recommend to disable `no-unused-vars` rule because TypeScript compiler itself verifies it.

ficristo:
I tryed to use `eslint-plugin-typescript` without success: https://github.com/bradzacher/eslint-plugin-typescript/issues/171
In the end I disabled the rule for ts and tsx files.
[...]

Issue #552, opened 2018-11-16
no-shadow if const and enum value are the same


samijaber:
**What version of TypeScript are you using?**
3.1.6

**What version of `typescript-eslint-parser` are you using?**
21.0.1

**What code were you trying to parse?**

```ts
const f = 0;
enum Foo {
  f,
}
```

**What did you expect to happen?**
No errors

**What happened?**
Got a `no-shadow` error

Might be somewhat related to #459

mysticatea:
Thank you for this report.

However, this is an expected behavior because the enum member `f` literally shadows the upper `f`.

```ts
const f = 777;
enum E1 {
    f = 1, // Shadows the upper `f`.
    g = f, // = 1
}
enum E2 {
    f1 = 1,
    g = f, // This is the upper `f`.
}
```

> [Open playground](https://www.typescriptlang.org/play/index.html#src=const%20f%20%3D%20777%3B%0D%0Aenum%20E1%20%7B%0D%0A%20%20%20%20f%20%3D%201%2C%20%2F%2F%20Shadows%20the%20upper%20%60f%60.%0D%0A%20%20%20%20g%20%3D%20f%2C%20%2F%2F%20%3D%201%0D%0A%7D%0D%0Aenum%20E2%20%7B%0D%0A%20%20%20%20f1%20%3D%201%2C%0D%0A%20%20%20%20g%20%3D%20f%2C%20%2F%2F%20This%20is%20the%20upper%20%60f%60.%0D%0A%7D%0D%0A)

PR #558, opened 2018-11-21
Fix: make no-unused-vars not reporting enum members

mysticatea:
This is a small followup for #553.

Currently, `no-unused-vars` rule reports enum members.

```ts
enum E {
    FOO //→ 'FOO' is assigned a value but never used.
}
```

This behavior is not nice.

This PR makes to set `eslintUsed` flag to the enum member variables, as similar to `/* exported */` directive comments. As a result, `no-unused-vars` knows the enum members are exported (used on other places).

### DIFF
--- a/analyze-scope.js
+++ b/analyze-scope.js
@@ -498,6 +498,11 @@ class Referencer extends OriginalReferencer {
         const scope = this.currentScope();
 
         scope.__define(id, new Definition("EnumMemberName", id, node));
+
+        // Set `eslintUsed` flag to the defined variable because the enum member is obviously exported.
+        const variable = scope.set.get(id.name);
+        variable.eslintUsed = true;
+
         if (initializer) {
             scope.__referencing(
                 id,
--- a/tests/lib/__snapshots__/scope-analysis.js.snap
+++ b/tests/lib/__snapshots__/scope-analysis.js.snap
@@ -2441,7 +2441,7 @@ Object {
               "type": "EnumMemberName",
             },
           ],
-          "eslintUsed": undefined,
+          "eslintUsed": true,
           "identifiers": Array [
             Object {
               "name": "A",
@@ -2488,7 +2488,7 @@ Object {
               "type": "EnumMemberName",
             },
           ],
-          "eslintUsed": undefined,
+          "eslintUsed": true,
           "identifiers": Array [
             Object {
               "name": "B",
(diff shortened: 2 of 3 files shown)


platinumazure:
I wonder if this is the right approach?

My thinking: Enums are sort of like objects, and enum values are sort of like object properties. I don't think we check object properties in no-unused-vars. So I feel that enum values also shouldn't be considered variables and shouldn't be added to a variable scope.

Maybe this isn't possible for other reasons? Let me know.

mysticatea:
It's better if the enum members behaved like properties. In fact, those close to variables. See also: https://github.com/eslint/typescript-eslint-parser/issues/552#issuecomment-439276282

armano2:
i'm unsure if this is good way to do it, it will be better if this will be supported by eslint-plugin-typescript ?

mysticatea:
I think that this approach is fair enough because:

1. The enum members are exported from the scope of the enum body. 
2. This is the very similar issue to that global variables are exported from the file.
3. We handle the exported variables by the `eslintUsed` property.

mysticatea:
The `typescript/use-enum-members` rule or something like (as similar to [react/jsx-uses-vars](https://github.com/yannickcr/eslint-plugin-react/blob/a33bfb18ca515e824c5ade55aab1505877b4501e/lib/rules/jsx-uses-vars.js)) work for me. 
But I think nice if it's supported in the parser because the similar `/* exported */` directive is supported in core.

eventualbuddha:
This issue is causing my team to sprinkle `/* eslint-disable no-unused-vars */` in our codebase whenever we have an enum. I'm happy to help fix this, or to augment `typescript/no-unused-vars` to handle this. What do y'all think should be done here?

JamesHenry:
@armano2 @bradzacher Will merging this PR interfere with the plugin's implementation in any way?

bradzacher:
@JamesHenry - I don't think so. In fact, it means we can delete some code!

https://github.com/bradzacher/eslint-plugin-typescript/blob/c9d2227db2bea041bcd3dc21c639b05b0732d61e/lib/rules/no-unused-vars.js#L63-L65

kaicataldo:
Closing this PR since the project has been moved to the [TypeScript ESLint organization](https://typescript-eslint.io). Feel free to reopen the PR there.

PR #560, opened 2018-11-21
Fix: order of visiting in TSAbstractClassProperty

armano2:


armano2:
@platinumazure can i have some info when those fixes can be merged?

PR #561, opened 2018-11-21
Fix: visiting superTypeParameters in classes

armano2:
This PR is adding `superTypeParameters` to class like types `ClassDeclaration`, `ClassExpression`, `TSAbstractClassDeclaration` and fixes order of `TSAbstractClassDeclaration`


armano2:
@platinumazure should i put each one of missing visitor key in each own PR or should i make one big PR?

platinumazure:
@armano2 Up to you. Please spare a thought for the poor reviewer that has to review a 1000+ line PR :smile: But if it's not that big, you could do one or a few PRs rather than one per node type.

armano2:
its more than 1k line with tests xd

platinumazure:
@armano2 I would say do what is most convenient for you and worst case, we can ask for changes. I'm not super worried-- I trust you will have good judgment. (If there are areas of code where collisions would happen, let's do those in combined PRs to make things easier.)

tests/fixtures/scope-analysis/class-supper-type.ts
```diff
@@ -0,0 +1,11 @@
+abstract class Foo extends Bar<Baz> {
```
mysticatea: I'm guessing the `Bar<Baz>` should be a `TSTypeReference` node...

armano2: `<Baz>` is TSTypeReference

armano2: https://gist.github.com/armano2/b26b9e77438c72c5b1c875024272c2ed

armano2: superTypeParameters -> TSTypeParameterInstantiation[params] -> TSTypeReference

armano2: `Bar` is `superClass` -> `Identifier`

mysticatea: Mmm. It's not ideal...

```ts
class Foo1 extends T<U> {}
class Foo2 extends Wrap<T<U>> {}
function f(a: T<U>): T<U> {}
const A: T<U> = {}
```

Every `T<U>` should be:

```jsonc
{
  "type": "TSTypeReference", // T<U>
  "typeName": {
    "type": "Identifier", // T
    "name": "T"
  },
  "typeParameters": {
    "type": "TSTypeParameterInstantiation", // <U>
    "params": [
      {
        "type": "TSTypeReference", // U
        "typeName": {
          "type": "Identifier",
          "name": "U"
        }
      }
    ]
  }
```

But the first one is different...

mysticatea: Ah... I see, because the `T` of the `extends T` is a variable rather than a type.
Mmmm.

mysticatea: OK, fair enough.

armano2: `extends T` is in superClass

platinumazure:
@armano2 Merged, thanks for contributing!

PR #562, opened 2018-11-22
Fix: visiting implements in classes

armano2:
This PR adds visiting of `implements` to classes


analyze-scope.js
```diff
@@ -210,11 +210,19 @@ class Referencer extends OriginalReferencer {
...
      * Visit decorators.
-     * @param {ClassDeclaration|ClassExpression} node The class node to visit.
+     * @param {ClassDeclaration|ClassExpression|TSAbstractClassDeclaration} node The class node to visit.
      * @returns {void}
      */
     visitClass(node) {
         this.visitDecorators(node.decorators);
+
+        const upperTypeMode = this.typeMode;
+        this.typeMode = true;
```
platinumazure: I think it might be better to move this line (and `this.typeMode = upperTypeMode`) into the body of the `if` statement. What do you think?

armano2: not really cus in 2nd PR i added one more case in there for supperClassType

armano2:
ok, now i have to just resolve conflicts here :>

analyze-scope.js
platinumazure: I just don't understand the value of setting `this.typeMode` to true (and back to its previous value) if the `if` statement evaluates to false. Feels like wasted cycles. Am I missing something?

platinumazure: Oh, now I see-- implements is also covered here. Okay, this is fine as is. Thanks!

armano2: if `typeMode` is set to true, we are working in diffrent scope of variables (type variables) from typescript

and we don't want to register type variables as js variables for eslint, there is some cases like here

https://github.com/eslint/typescript-eslint-parser/pull/562/files/05bf00af0cca3e43ca566c99c436a277bebfe23a#diff-32f897025e0e51a215a578a256455ab5L230

that we are not using eslint implementation of node if we are in "type scope"

armano2: i hope i did not make it more complex than it is xd

platinumazure: It's fine, I just didn't realize that `implements` was also working with type variables. Without the `implements` section, I think it would have made sense to put the typeMode change inside the `if` body like you did in another PR. But in this case we need to cover both `superTypeParameters` and `implements`.

platinumazure:
@armano2 Merged, thanks for contributing!

armano2:
@platinumazure thank you for unblocking migration of [`eslint-plugin-typescript`](https://github.com/bradzacher/eslint-plugin-typescript/pull/174) to latest version

Issue #555, opened 2018-11-20
Missing visitor-keys


armano2:
I did some checking what we can and what we are not able to visit and i found some missing stuff:

#### missing `typeParameters`
- [x] NewExpression - `new foo<Foo>()` #565
- [x] CallExpression - `foo<Foo>()` #565
- [x] TSAbstractClassDeclaration:  #561 - `abstract class Foo<Bar> {}`

#### missing `superTypeParameters` 
`class foo extends bar<Foo>`
- [x] ClassDeclaration: #561
- [x] ClassExpression: #561
- [x] TSAbstractClassDeclaration: #561

#### missing `implements`
`class foo implements bar`,  `class foo implements bar, baz` 
- [x] ClassDeclaration: #562
- [x] ClassExpression: #562
- [x] TSAbstractClassDeclaration: #562

#### missing import equals
`import foo = require('foo')`
- [x] TSImportEqualsDeclaration #571
  - name
  - moduleReference
- [x] TSExternalModuleReference #571
  - expression

#### invalid order:
- [x] TSAbstractClassProperty : #560
  - key
  - typeAnnotation
  - value

#### missing decorators:
- [x] TSAbstractClassProperty: #560
- [x] TSAbstractClassDeclaration #561

armano2:
small update

mysticatea:
No. 

TypeScript has two namespaces: variables and types.

```ts
// Those are not redeclaration 
// because `interface` defines only in types and `const` defines only in variables.
interface A {}
const A = 0

// On the other hand, `class` defines in both types and variables.
class B {}
```

Then, the `scopeManager` doesn't have types.
I think that we should provide a way to access `types` namespace, but we don't have it now.

armano2:
ok, i was thinking correctly,

i just wanted to confirm that :)

PR #571, opened 2018-11-28
Fix: scope for TSImportEqualsDeclaration

armano2:
Import equal should declare "ImportBinding" variable in scope.

example:
```ts
import foo = require('bar')
```

https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

this PR fixes ticket #555


tests/lib/__snapshots__/scope-analysis.js.snap
```diff
@@ -4480,6 +4480,101 @@ Object {
...
+              "name": "foo",
+              "range": Array [
+                7,
+                10,
+              ],
+              "type": "Identifier",
+            },
+          ],
+          "name": "foo",
+          "references": Array [],
```
mysticatea: In TS, isn't the `require` a reference?

j-f1: No — `import foo = require('bar')` is different from `const foo = require('bar')

armano2: i will investigate that

armano2: @mysticatea  `import foo = require('bar')` is different syntax to write `import foo from 'bar'`
and `require` is not reference like `from` is not.

typescript have this weird syntax to distinct `export =` from `export default`

mysticatea: Thank you for the explanation! I got it.

Issue #570, opened 2018-11-28
declare global causing parse errors


milesj:
**What version of TypeScript are you using?**

3.0

**What version of `typescript-eslint-parser` are you using?**

21.0.1

**What code were you trying to parse?**

```ts
declare global {
  namespace NodeJS {
    interface Process {
      beemo: {
        context: any;
        tool: BeemoTool;
      };
    }
  }
}
```

**What did you expect to happen?**

It parse.

**What happened?**

It fails with this error.

```
/Users/Miles/Sites/beemo/packages/core/src/index.ts
  17:15  error  Parse errors in imported module './types': Cannot read property '0' of undefined (undefined:undefined)  import/export
```

This has happened on 2 projects now, both because of `declare global`.

milesj:
Probably more related to `export *` coupled with `declare global`.

Here's the index file: https://github.com/milesj/beemo/blob/master/packages/core/src/index.ts#L17

And types file: https://github.com/milesj/beemo/blob/master/packages/core/src/types.ts#L62

armano2:
@milesj this issue is in https://github.com/benmosher/eslint-plugin-import 

this plugin is not supporting typescript https://github.com/benmosher/eslint-plugin-import/issues?q=is%3Aopen+is%3Aissue+label%3Atypescript

milesj:
This was working before a recent parser upgrade though. Do you know exactly what changed that may have caused this?

armano2:
@milesj hmm i did some testing and i set up config and all without your wrappers and its working fine

milesj:
This is the eslint config I'm using: https://github.com/milesj/build-tool-config/blob/master/packages/config/configs/eslint.js

Nothing atm stands out.

armano2:
yea i found it, and looks like its issue in AST after all xd

armano2:
```
TypeError: Cannot read property '0' of undefined
    at isGlobalAugmentation (******\typescript-eslint-parser\analyze-scope.js:32:26)
    at Referencer.TSModuleDeclaration (******\typescript-eslint-parser\analyze-scope.js:578:13)
    at Referencer.Visitor.visit (******\typescript-eslint-parser\node_modules\esrecurse\esrecurse.js:104:34)
    at Referencer.Visitor.visitChildren (******\typescript-eslint-parser\node_modules\esrecurse\esrecurse.js:83:38)
[...]

PR #567, opened 2018-11-25
Docs: Update README.md

dudiharush:
Typescript support is 3.1.1 according to package.json file

PR #574, opened 2018-11-30
Fix: issue with global augmentation when tokens are not present

armano2:
This PR solves issue with: https://github.com/eslint/typescript-eslint-parser/issues/570

i had to upgrade `typescript-estree` to latest version with solution for https://github.com/JamesHenry/typescript-estree/issues/27

additionally lodash is no longer used in project


armano2:
@mysticatea you should look on this one https://github.com/eslint/typescript-eslint-parser/pull/567

PR #581, opened 2018-12-16
Fix: Allow to visit typeParameters in VariableDeclarator

armano2:
This PR adds `typeParameters` to VariableDeclarator

`typeParameters` can be only present if `kind=type`

### DIFF
--- a/visitor-keys.js
+++ b/visitor-keys.js
@@ -21,6 +21,8 @@ module.exports = Evk.unionWith({
     ObjectPattern: ["properties", "typeAnnotation"],
     NewExpression: ["callee", "typeParameters", "arguments"],
     CallExpression: ["callee", "typeParameters", "arguments"],
+    // typeParameters are preset only when kind='type'
+    VariableDeclarator: ["id", "typeParameters", "init"],
 
     // Additional Nodes.
     ClassProperty: ["decorators", "key", "typeAnnotation", "value"],
--- a/visitor-keys.js
+++ b/visitor-keys.js
@@ -21,7 +21,7 @@ module.exports = Evk.unionWith({
     ObjectPattern: ["properties", "typeAnnotation"],
     NewExpression: ["callee", "typeParameters", "arguments"],
     CallExpression: ["callee", "typeParameters", "arguments"],
-    // typeParameters are preset only when kind='type'
+    // typeParameters are present only when kind='type'
     VariableDeclarator: ["id", "typeParameters", "init"],
 
     // Additional Nodes.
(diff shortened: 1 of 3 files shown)


visitor-keys.js
```diff
@@ -21,6 +21,8 @@ module.exports = Evk.unionWith({
     ObjectPattern: ["properties", "typeAnnotation"],
     NewExpression: ["callee", "typeParameters", "arguments"],
     CallExpression: ["callee", "typeParameters", "arguments"],
+    // typeParameters are preset only when kind='type'
```
platinumazure: Typo: preset --> present

armano2:
@platinumazure fixed

visitor-keys.js
armano2: @platinumazure fixed

armano2:
@platinumazure can i have estimate when this change can be released? its blocking me from fixing some of rules.

platinumazure:
@armano2 Very sorry I missed your comment.

Is this PR still needed given the changes in #589 (i.e., is that PR separate/independent from this one)? Thanks!

armano2:
this PR is needed only for estree v6 in v7 there was change that 
```ts
type Foo = number
```
is no longer variable and now its its own node `TSTypeAliasDeclaration`
https://github.com/eslint/typescript-eslint-parser/pull/589/commits/6aad31d5edf06d25677651446d9a88fdd29ef990#diff-32f897025e0e51a215a578a256455ab5

platinumazure:
Thanks, makes sense.

In the PRs for upgrading typescript-estree to 7.x or higher, do we need to go back to visitor-keys.js and remove the override for variable declarations, given the decision to switch the node type for `type` declarations?

armano2:
this property is no longer in AST, and it will be better if it's going to be removed when we upgrade to 7.x

platinumazure:
@armano2 Understood. Sorry, I wasn't clear: If I merge this PR now, do you need to update #584 and #589 to ensure the property is removed from visitor-keys?

If so, then, given this extra work needed on your side: Should I merge this PR at all (and cut a patch release)? Or should we just close this and have people migrate to a new major version of this plugin (one which supports typescript-estree@7.x or typescript-estree@8.x)?

I'm leaning towards just closing this, but if you think there's a strong need to fix this for people stuck on earlier versions of TS, I could merge this and do a release, and then request changes on #584 and #589. Let me know what you think the best way forward is here.

armano2:
i don't mind rebasing / removing it there

armano2:
any progress on this?

JamesHenry:
@armano2 We can close this in favour of #596, right? The node kind in question is no longer available

Issue #414, opened 2017-12-13
`type` declarations should have distinct AST from `var`s


JamesHenry:
- We currently generate a "special" variable declaration for a type declaration.
- This was done to allow core rules such as `semi` to work with type declarations out of the box.
- However, other core rules (and plugin rules) have issues with this

It leads to issues such as https://github.com/eslint/typescript-eslint-parser/issues/350 https://github.com/eslint/typescript-eslint-parser/issues/359

We should come up with a distinct AST for type declarations and figure out how to make `semi` work in some other way.

IanEdington:
This is throwing `no-restricted-globals`:

```
import * as React from 'react';

import { Name } from './styled';

export interface propTypes {
  name: string;
}

export default class SomeClass extends React.PureComponent<propTypes> {
  render() {
    const { name } = this.props;

    return <Name>{name}</Name>;
  }
}
```

majelbstoat:
```typescript
export default class MyComponent extends React.Component<Props, State> {
  open = () => {
    console.log('something')
  }
}
```

This is also throwing `no-restricted-globals` for `open`.

billyvg:
Is this related to the following throwing `no-eval` errors?
```
export type Spec = {
    eval?: string;
};
```

PR #576, opened 2018-12-06
Breaking: Bump to typescript-estree@6.0.0, typescript@~3.2.1

nevir:
From what I can tell, the main change in typescript-estree is added support for `BigIntLiteral`s. Added some snapshots for 'em, as well.

---

Should this be a major bump due to the typescript upgrade? Or only minor?

Issue #384, opened 2017-09-10
`declare class` is represented the same as a regular class


j-f1:
**What version of TypeScript are you using?** 2.4.2

**What version of `typescript-eslint-parser` are you using?** 6.0.1

**What code were you trying to parse?**

```ts
declare class Foo {}
declare var Foo
declare function Foo() {}
```

**What did you expect to happen?**

The AST should specify that the `Foo` class and the `Foo` variable aren’t actual declarations.

Maybe

```ts
// declare var Foo
({
  type: 'VariableDeclaration',
  declarations: 
   [ { type: 'VariableDeclarator',
       id: 
        { type: 'Identifier',
         name: 'Foo' },
      init: null } ],
  kind: 'declare var'
})

// declare class Foo {}
({
  type: 'TSAmbientClass',
  id: 
   { type: 'Identifier',
     name: 'Foo' },
  body: 
   { type: 'ClassBody',
     body: [],
  superClass: null
})
```

**What happened?**

The AST is the same as the one generated for this code:

```ts
class Foo {}
var Foo
declare function Foo() {}
```

Generated AST:

```ts
({ type: 'Program',
  body: 
   [ { type: 'ClassDeclaration',
       id: 
        { type: 'Identifier',
          name: 'Foo' },
       body: 
        { type: 'ClassBody',
          body: [] },
       superClass: null },
     { type: 'VariableDeclaration',
       declarations: 
        [ { type: 'VariableDeclarator',
            id: 
             { type: 'Identifier',
               name: 'Foo' },
            init: null } ],
       kind: 'var' },
     { type: 'DeclareFunction',
       id: 
        { type: 'Identifier',
          name: 'Foo' },
       generator: false,
       expression: false,
       async: false,
       params: [],
       body: 
        { type: 'BlockStatement',
          body: [] } } ],
  sourceType: 'script' })
```

j-f1:
This breaks `no-useless-constructor`.

mysticatea:
I removed `ESLint rule/integration issue` label because this is an AST issue that the AST cannot distinguish semantics. So let's solve it on the underlying package.

/cc @JamesHenry

JamesHenry:
Fxed by #596

PR #584, opened 2018-12-24
Breaking: typescript-estree to 7.0.0 and typescript to 3.2.1

armano2:
This PR contains changes from #576 and deprecates changes from #581 

recently i started fixing issues in `typescript-estree` (most of changes are improvements to typing and tests)

but there was important changes to generated AST:

- https://github.com/JamesHenry/typescript-estree/pull/48
- https://github.com/JamesHenry/typescript-estree/pull/46
- https://github.com/JamesHenry/typescript-estree/pull/42

------------

- There was change of one node: `DeclareFunction` got renamed to `TSDeclareFunction` to remove confuction.
- Alot of fields got update and we have now optional `declare` property to determine if node is "declaration". `TSDeclareFunction` has `declare` property always set to true.
- `VariableDeclarator[kind='type']` was causing a lot of issues within esling rules and it was extracted to its own type: `TSTypeAliasDeclaration`
- modifier `abstract` is now optional (`true` or `undefined`)

------

closes: #581, #576 
fixes: #414, #588, #384


platinumazure:
Hi @armano2, should this be closed now that #589 has been created?

armano2:
depends how you want to release it, if you prefer doing smaller patches, you can release them one by one. and i will rebase changes.

platinumazure:
@armano2 I don't see much value in releasing with typescript-estree 7.x unless there is a good chance that there are people in the community who are using typescript-estree 7.x for their projects, are currently blocked from using ESLint since we're still on a 5.x version, and would also be blocked from upgrading if we went straight to 8.x. If there isn't that explicit need to provide some support for 7.x, I would rather just jump straight to 8.x since it's a bit less work on the maintenance side.

@kaicataldo @mysticatea What do you think of the above? Should we cut a major release of typescript-eslint-parser with typescript-estree 7.x support "just in case", or can we jump to 8.x? And if we want to support 7.x as well, should we do a release with 6.x before that?

JamesHenry:
@platinumazure I'm sorry for the noise - it's purely down to this project and typescript-estree having very different approaches to releases.

I wanted typescript-estree to be a low touch as possible - there is a strict following of semver, and semantic-release handles everything automatically upon merge to master.

I would strongly suggest the "major" version of TypeScript is used as the source of truth (TypeScript does not follow semver, so this is not all that intuitive, but major in terms of TypeScript effectively refers to the second digit. E.g. the parser would move from supporting X.1 to X.2)

Please let me know if you want to discuss it any further.

For these PRs specifically, I see no reason not to jump straight to the latest typescript-estree, as the version of TypeScript which is targeted/supported has not changed. It's just that we have been refining the AST structure a lot recently, and all changes are released as individual major version bumps.

I am doing a similar (on the face of it) large jump in Prettier right now: https://github.com/prettier/prettier/pull/5728 - as you can see the amount of changes needed are not as intense as you might have initially expected from v6 -> v13

kaicataldo:
As far as upgrading `typescript-eslint` goes, I like the idea of tracking the supported version of TypeScript (following the version of semver that TS follows as described by @JamesHenry).

JamesHenry:
Closing in favour of https://github.com/eslint/typescript-eslint-parser/pull/596

PR #595, opened 2019-01-09
Fix: support ecmaFeatures.jsx flag

### LINKED ISSUE #594: parserOptions.ecmaFeatures.jsx option not passed to estree
3.1.1

21.0.2

```js
const RuleTester = require("eslint").RuleTester;

const ruleTester = new RuleTester({
    parserOptions: {
        ecmaVersion: 6,
        sourceType: "module",
        ecmaFeatures: {
            jsx: true,
        },
    },
    parser: "typescript-eslint-parser",
});

ruleTester.run("no-unused-vars", rule, {
    valid: [
		`
const Foo = function () {}
function render() {
    return (<Foo />);
}
		`,
	],
	invalid: [],
})
```

The code should parse successfully, so I can test the rule.

**What happened?**
`Parsing error: \'>\' expected.`

[see astexplorer repl](https://astexplorer.net/#/gist/45a09d9c2abdf3ae1d59e6b703d05913/d8b1c7667fe0549b931abbcc3078c8a665d5f8c4)
Note the parser has the `jsx` option set to true

bradzacher:
Fixes #594

### DIFF
--- a/parser.js
+++ b/parser.js
@@ -32,7 +32,7 @@ exports.parseForESLint = function parseForESLint(code, options) {
             options = Object.assign({}, options, { jsx: tsx });
         }
     }
-    if (options.ecmaFeatures && options.ecmaFeatures.jsx) {
+    if (options.ecmaFeatures && typeof options.ecmaFeatures.jsx === "boolean") {
         // allow the user to override the jsx setting
         options.jsx = options.ecmaFeatures.jsx
     }
--- a/parser.js
+++ b/parser.js
@@ -32,6 +32,10 @@ exports.parseForESLint = function parseForESLint(code, options) {
             options = Object.assign({}, options, { jsx: tsx });
         }
     }
+    if (options.ecmaFeatures && options.ecmaFeatures.jsx) {
+        // allow the user to override the jsx setting
+        options.jsx = options.ecmaFeatures.jsx
+    }
 
     const ast = parse(code, options);
     const extraOptions = {
--- a/parser.js
+++ b/parser.js
@@ -32,7 +32,9 @@ exports.parseForESLint = function parseForESLint(code, options) {
             options = Object.assign({}, options, { jsx: tsx });
         }
     }
-    if (options.ecmaFeatures && typeof options.ecmaFeatures.jsx === "boolean") {
+    if (typeof options.jsx !== "boolean" &&
+        options.ecmaFeatures &&
+        typeof options.ecmaFeatures.jsx === "boolean") {
         // allow the user to override the jsx setting
         options.jsx = options.ecmaFeatures.jsx;
     }
--- a/parser.js
+++ b/parser.js
@@ -33,7 +33,7 @@ exports.parseForESLint = function parseForESLint(code, inputOptions) {
 
     options.jsx = (options.ecmaFeatures || {}).jsx;
     if (typeof options.jsx !== "boolean") {
-        inputOptions.jsx = false;
+        options.jsx = false;
     }
 
     // override the jsx option depending on the file path
--- a/parser.js
+++ b/parser.js
@@ -34,7 +34,7 @@ exports.parseForESLint = function parseForESLint(code, options) {
     }
     if (options.ecmaFeatures && typeof options.ecmaFeatures.jsx === "boolean") {
         // allow the user to override the jsx setting
-        options.jsx = options.ecmaFeatures.jsx
+        options.jsx = options.ecmaFeatures.jsx;
     }
 
     const ast = parse(code, options);
--- a/parser.js
+++ b/parser.js
@@ -20,23 +20,28 @@ const visitorKeys = require("./visitor-keys");
 
 exports.version = require("./package.json").version;
 
-exports.parseForESLint = function parseForESLint(code, options) {
-    if (typeof options !== "object" || options === null) {
-        options = { useJSXTextNode: true };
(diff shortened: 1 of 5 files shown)


parser.js
```diff
@@ -32,6 +32,10 @@ exports.parseForESLint = function parseForESLint(code, options) {
             options = Object.assign({}, options, { jsx: tsx });
         }
     }
+    if (options.ecmaFeatures && typeof options.ecmaFeatures.jsx === "boolean") {
```
armano2: can you check if `options.jsx` is not defined already here?

we don't want to change behavior for ppl who have some options already set.

kaicataldo:
I'm not a big fan of having two ways to do the same thing. Any chance we can decide on one way to do this and be consistent about it (deprecating the old way if we decide on `parserOptions.ecmaFeatures.jsx`)?

My vote would be to make it consistent with the default parser and make `parserOptions.ecmaFeatures.jsx` the canonical way to set this and to deprecate `parserOptions.jsx`. babel-eslint [does this](https://github.com/babel/babel-eslint#additional-parser-configuration) with `ecmaFeatures.globalReturn` as well.

armano2:
@kaicataldo we should support only one case, but we can't do breaking changes without major release

platinumazure:
We'll be doing a major release with all the syntax changes in the typescript-estree upgrades @armano2, don't worry about that. Next release will probably be major. :smile:

armano2:
this could be done as hotfix -> and it will fix issues for users using configs like react within this version :>

bradzacher:
So what was the goal for this?
Would you like me to refactor to remove `options.jsx`, or is this good as is?

I'm happy to raise a second PR so that we can release both a patch and a breaking fix?

armano2:
i think we should merge this first, do patch release, and wait for "eslint team" to decide how they are going to handle releases for new version of estree

JamesHenry:
@bradzacher I'm back to reviewing things on this side of the fence again :)

I will be doing a major release of the library as soon as we merge https://github.com/eslint/typescript-eslint-parser/pull/596, so let's go for what we've deemed the most correct solution here.

bradzacher:
I have dropped support for inputting `options.jsx`

parser.js
```diff
@@ -20,16 +20,27 @@ const visitorKeys = require("./visitor-keys");
...
+        options.useJSXTextNode = true;
     }
-    if (typeof options.filePath === "string") {
-        const tsx = options.filePath.endsWith(".tsx");
-        if (tsx || options.filePath.endsWith(".ts")) {
-            options = Object.assign({}, options, { jsx: tsx });
+
+    options.jsx = (options.ecmaFeatures || {}).jsx;
+    if (typeof options.jsx !== "boolean") {
+        inputOptions.jsx = false;
```
armano2: `options.jsx` ?

bradzacher: oops copy pasta

parser.js
```diff
@@ -20,16 +20,27 @@ const visitorKeys = require("./visitor-keys");
...
-        if (tsx || options.filePath.endsWith(".ts")) {
-            options = Object.assign({}, options, { jsx: tsx });
+
+    options.jsx = (options.ecmaFeatures || {}).jsx;
+    if (typeof options.jsx !== "boolean") {
+        options.jsx = false;
+    }
+
+    // override the jsx option depending on the file path
+    if (typeof inputOptions.filePath === "string") {
```
kaicataldo: I'm not convinced this is the behavior we want. Since the parser currently sets this automatically based on file extension, this option should only be necessary when no file path is present. As such, autodetection should not happen when this option is set.

armano2: there is few examples of cases with no filePath:

vue files provided by `vue-eslint-parser`,
files provided by `eslint-html-parser`, `eslint-plugin-html`, 
and potentially files provided by https://github.com/eslint/eslint-plugin-markdown in feature..

bradzacher: I was just going by the behaviour listed in the readme:

- It's `false` on `*.ts` files _regardless of this option_.
- It's `true` on `*.tsx` files _regardless of this option_.

armano2: yes but this is done by typescript itself https://www.typescriptlang.org/docs/handbook/jsx.html

> In order to use JSX you must do two things.
>
> * Name your files with a .tsx extension
> * Enable the jsx option

JamesHenry:
Sorry for the conflicts, @armano2 is your "approved" review status still valid on this?

armano2:
there is only think think pointed out by @kaicataldo and i'm unsure what to do with this...

kaicataldo:
Sorry, I didn’t realize that this was already the current behavior. I’ll resolve my comment!

parser.js
kaicataldo: Sorry, I didn’t realize this was the current behavior.

JamesHenry:
If it's ok with @bradzacher (he's on the core team for the new typescript-eslint org) I would rather get https://github.com/typescript-eslint/typescript-eslint/pull/6 over the line and press on from there.

We can quickly put this and the other PRs together on that repo

bradzacher:
Happy with whatever is easier. This is a simple pr to remake on the new repo :)

kaicataldo:
Closing this PR since the project has been moved to the [TypeScript ESLint organization](https://typescript-eslint.io). Feel free to reopen the PR there.

